PR #231, opened 2024-08-28
Allow shorthand for object input for functions with user / assistant input schemas

### LINKED ISSUE #211: Allowing JSON objects in the shorthand we allow for text
Currently, in order to call a function with an input schema using the Python client, we need to do the following:

```python
 haiku_result = await client.inference(
        function_name="write_haiku",
        input={
            "messages": [
                {
                    "role": "user",
                    "content": [{"type": "text", "value": {"topic": topic}}],
                }
            ]
        },
    )
```
This requires 2 potentially unnecessary layers of JSON and also a "type": "text" field.

If we did a similar function today that used a string message, we would be able to do this:

```python
 await client.inference(
        function_name="write_haiku",
        input={
            "messages": [
                {
                    "role": "user",
                    "content": "hello!",
                }
            ]
        },
    )
```

I propose that we allow a similar syntax for JSON input messages that are Objects as well as the current behavior that allows this with strings: 
```python
 haiku_result = await client.inference(
        function_name="write_haiku",
        input={
            "messages": [
                {
                    "role": "user",
                    "content":  {"topic": topic},
                }
            ]
        },
    )
```
I still think that less common message types such as tool calls / results should require the long-form input as they currently do, and any message with more than one block will obviously also require it.
Potential issues: 2 ways of doing the same thing.

virajmehta:
Closes #211 

----

> [!IMPORTANT]
> Add shorthand for object input in functions with user/assistant input schemas, update deserialization logic, and enhance tests and schemas.
> 
>   - **Behavior**:
>     - Allow shorthand for object input in functions with user/assistant input schemas by updating `Deserialize` implementation in `types.rs`.
>     - Change `Error::JsonSchemaValidation` to return `BAD_REQUEST` in `error.rs`.
>   - **Tests**:
>     - Update `test_client.py` and `inference.rs` to test object input handling.
>     - Add `test_deserialize_input_message()` in `types.rs` to cover new input cases.
>   - **Schemas and Templates**:
>     - Add `user_schema.json` and `user_template.minijinja` for `json_success` function.
>     - Update `tensorzero.toml` to include new schemas and templates for `json_success` variants.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 97181458a168dacd43938f390333968da7e337a7. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/clients/python-asyncio/tests/test_client.py
+++ b/clients/python-asyncio/tests/test_client.py
@@ -221,7 +221,7 @@ async def test_json_streaming(client):
         function_name="json_success",
         input={
             "system": {"assistant_name": "Alfred Pennyworth"},
-            "messages": [{"role": "user", "content": "Hello, world!"}],
+            "messages": [{"role": "user", "content": {"country": "Japan"}}],
         },
         stream=True,
     )
@@ -263,12 +263,12 @@ async def test_json_streaming(client):
 
 
 @pytest.mark.asyncio
-async def test_json_sucess(client):
+async def test_json_success(client):
     result = await client.inference(
         function_name="json_success",
         input={
             "system": {"assistant_name": "Alfred Pennyworth"},
-            "messages": [{"role": "user", "content": "Hello, world!"}],
+            "messages": [{"role": "user", "content": {"country": "Japan"}}],
         },
         stream=False,
     )


GabrielBianconi:
@ellipsis-dev review [PR summary]

PR #309, opened 2024-09-10
Added support for dynamic output schemas in JSON functions

virajmehta:
This PR involved changing the ModelInferenceRequest.tool_config to a Cow in order to handle that usually ToolConfigs are generated at request time or statically at startup but can be generated at variant inference time because we have the possibility of a dynamic JSON out schema getting routed to an implicit tool variant. 

Since we changed this to a Cow we had to add Arcs to the JSON schemas in tools throughout the code in order to allow copies. We don't actually use this clone-on-write functionality since the only writes would be to the owned variant (joining the task). 

Option for a different design: just clone everywhere since the JSON schemas are Arced now. I think that would still be heavier and worse than having a Cow type.


gateway/src/endpoints/inference.rs
```diff
@@ -59,6 +61,7 @@ pub struct Params {
     // tool_choice: Option<ToolChoice>,
     // If true, the inference will use parallel tool calls
     // parallel_tool_calls: Option<bool>,
+    output_schema: Option<Value>,
```
GabrielBianconi, line 64: document

gateway/src/function.rs
```diff
@@ -167,23 +164,30 @@ impl FunctionConfig {
...
                         Ok(_) => Some(parsed_output),
                         Err(e) => {
                             e.log();
                             None
                         }
                     }
-                });
+                } else {
+                    None
+                };
```
GabrielBianconi, line 183: Use an async closure

gateway/fixtures/config/functions/dynamic_json/output_schema.json
```diff
@@ -0,0 +1 @@
+{}
```
GabrielBianconi, line 1: slop alert

GabrielBianconi (changes requested):

gateway/src/config_parser.rs
```diff
@@ -859,7 +859,7 @@ mod tests {
             .contains("Failed to parse config:\nunknown field `enable_agi`, expected"));
     }
 
-    /// Ensure that the config parsing fails when a JSON function has no output schema
+    /// Ensure that the config parsing defaults properly for JSON functions with no output schema
     #[test]
```
GabrielBianconi, line 863: We need a unit test somewhere that checks the default output schema works as expected (i.e. accepts any JSON and rejects everything else).

GabrielBianconi (changes requested):

GabrielBianconi (approved):

PR #315, opened 2024-09-10
Validates there are credentials for models used by variants with positive weight

### LINKED ISSUE #224: Gateway should fail on startup if there are variants with weight that use API keys that haven't been provided
Should also consider how this interacts with API keys provided at request time

virajmehta:
Closes #224

### DIFF
--- a/gateway/src/inference/providers/openai.rs
+++ b/gateway/src/inference/providers/openai.rs
@@ -120,6 +120,10 @@ impl InferenceProvider for OpenAIProvider {
         };
         Ok((chunk, stream))
     }
+
+    fn has_credentials(&self) -> bool {
+        self.api_key.is_some()
+    }
 }
 
 pub fn stream_openai(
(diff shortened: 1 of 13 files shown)


gateway/src/inference/providers/openai.rs
```diff
@@ -120,6 +120,10 @@ impl InferenceProvider for OpenAIProvider {
         };
         Ok((chunk, stream))
     }
+
+    fn has_credentials(&self) -> bool {
+        self.api_key.is_some()
+    }
```
GabrielBianconi, line 126: If you're using an unauthenticated OpenAI-compatible endpoint, this will break.

We should check ~ "API base is None or API key is Some"

GabrielBianconi (changes requested):

GabrielBianconi (approved):

GabrielBianconi:
@ellipsis-dev review [retroactively add rich PR description]

PR #324, opened 2024-09-14
Refactor variant into directory structure and put validation into individual configs

virajmehta:


gateway/src/model.rs
```diff
@@ -106,6 +106,20 @@ impl ModelConfig {
         }
         Err(Error::ModelProvidersExhausted { provider_errors })
     }
+
+    pub fn validate(&self) -> Result<(), Error> {
+        // Ensure that at least one provider has credentials
+        if !self
+            .providers
+            .values()
+            .any(|provider| provider.has_credentials())
```
GabrielBianconi, line 115: Maybe this should be a `all`? Otherwise we might accidentally always trigger a fallback unknowingly to the user

gateway/src/variant/chat_completion.rs
```diff
@@ -382,6 +236,111 @@ impl Variant for ChatCompletionConfig {
...
+    ///  - That the model name is a valid model
+    fn validate(
+        &self,
+        function: &FunctionConfig,
+        models: &HashMap<String, ModelConfig>,
+        templates: &TemplateConfig,
+        function_name: &str,
+        variant_name: &str,
+    ) -> Result<(), Error> {
+        // Check that the model name is a valid model
```
GabrielBianconi, line 254: Comment should move down

gateway/src/variant/chat_completion.rs
```diff
@@ -382,6 +236,111 @@ impl Variant for ChatCompletionConfig {
...
         Ok((first_chunk, Box::pin(stream), model_used_info))
     }
+
+    /// This function validates that the chat completion variant is correctly configured for the given function config.
+    /// In order to do this, we need to check:
+    ///  - For system, user, and assistant message types:
+    ///    - That the template here is provided if the schema is provided in the function.
+    ///    - If the template requires variables, the schema is provided.
+    ///  - That the model name is a valid model
+    fn validate(
```
GabrielBianconi, line 246: Add more inline comments / line breaks between logical blocks, tough to parse

GabrielBianconi (changes requested):
```
{"timestamp":"2024-09-15T21:19:57.843234Z","level":"ERROR","message":"Failed to load config: Error in TensorZero config: `functions.example_generate_draft.variants.openai_promptA.system_template`: Error in TensorZero config: template is required when schema is specified"}
```

Errors became very wordy / repetitive after this commit.

```
Error::Config { message } => {
                write!(f, "Error in TensorZero config: {}", message)
            }
```

Print `"{message}"` here instead. Top-level expect (`Failed to load config`) gives enough context.

gateway/src/model.rs
virajmehta: it will log, I think we should always do best-effort and in that case it would be a fallback that is in observability and logs both,

GabrielBianconi (approved):

PR #330, opened 2024-09-17
Implement small API changes for OpenAI o1 models

### LINKED ISSUE #325: Add support and testing for the O-1 series of models from OpenAI

virajmehta:
Closes #325

### DIFF
--- a/gateway/src/inference/providers/openai.rs
+++ b/gateway/src/inference/providers/openai.rs
@@ -25,6 +25,7 @@ lazy_static! {
         #[allow(clippy::expect_used)]
         Url::parse("https://api.openai.com/v1/").expect("Failed to parse OPENAI_DEFAULT_BASE_URL")
     };
+    static ref OPENAI_O1_MODELS: Vec<&'static str> = vec!["o1-preview", "o1-mini",];
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


gateway/src/inference/providers/openai.rs
```diff
@@ -96,10 +97,15 @@ impl InferenceProvider for OpenAIProvider {
         ),
         Error,
     > {
+        if OPENAI_O1_MODELS.contains(&self.model_name.as_str()) {
+            return Err(Error::InvalidRequest {
+                message: "o1 models do not support streaming".to_string(),
+            });
+        }
```
GabrielBianconi, line 104: Maybe we should check if `model_name` starts with `o1` and if so, categorize it as such. The next one is dropping soon and presumably will have the same limitations.

gateway/src/inference/providers/openai.rs
```diff
@@ -536,7 +552,46 @@ impl<'a> OpenAIRequest<'a> {
...
+            if let OpenAIRequestMessage::System(system_msg) = messages.remove(0) {
+                let user_msg = OpenAIRequestMessage::User(OpenAIUserRequestMessage {
+                    content: system_msg.content,
+                });
+                messages.insert(0, user_msg);
+            }
+        }
+        if request.tool_config.is_some() {
+            return Err(Error::InvalidRequest {
+                message: "O1 models do not support tools".to_string(),
```
GabrielBianconi, line 575: `The OpenAI o1 family of models does not support tools.`

gateway/src/inference/providers/openai.rs
```diff
@@ -96,10 +97,15 @@ impl InferenceProvider for OpenAIProvider {
         ),
         Error,
     > {
+        if OPENAI_O1_MODELS.contains(&self.model_name.as_str()) {
+            return Err(Error::InvalidRequest {
+                message: "o1 models do not support streaming".to_string(),
```
GabrielBianconi, line 102: `The OpenAI o1 family of models does not support streaming.`

GabrielBianconi (approved):

PR #359, opened 2024-09-27
together has added usage to their streaming API

virajmehta:


gateway/tests/e2e/providers/common.rs
```diff
@@ -6175,7 +6175,7 @@ pub async fn test_json_mode_streaming_inference_request_with_provider(provider:
     assert!(full_content.to_lowercase().contains("tokyo"));
 
     // NB: Azure and Together don't support input/output tokens during streaming
```
GabrielBianconi, line 6177: fix comment

virajmehta: kk

PR #361, opened 2024-09-28
Add experimental dynamic in-context learning variant type to gateway

virajmehta:

PR #363, opened 2024-09-29
Allows API keys to be passed at inference time

virajmehta:
Do this after #361 is merged


GabrielBianconi (changes requested):
No `_infer*`

gateway/src/endpoints/inference.rs
```diff
@@ -67,6 +68,8 @@ pub struct Params {
     // If provided for a JSON inference, the inference will use the specified output schema instead of the
     // configured one. We only lazily validate this schema.
     output_schema: Option<Value>,
+    #[serde(default)]
+    api_keys: InferenceApiKeys,
```
GabrielBianconi, line 72: credentials

gateway/src/endpoints/inference.rs
```diff
@@ -84,6 +87,20 @@ struct InferenceMetadata<'a> {
...
+    pub azure_openai_api_key: Option<SecretString>,
+    pub fireworks_api_key: Option<SecretString>,
+    pub gcp_vertex_api_key: Option<SecretString>,
+    pub mistral_api_key: Option<SecretString>,
+    pub openai_api_key: Option<SecretString>,
+    pub together_api_key: Option<SecretString>,
+    pub vllm_api_key: Option<SecretString>,
+    #[cfg(any(test, feature = "e2e_tests"))]
+    pub dummy_api_key: Option<SecretString>,
+}
```
GabrielBianconi, line 102: Double nest it: e.g. `openai.api_key` and so on.

Skip azure for now since it depends on deployment ID and other factors (to be implemented later.

This allows us to customize more parameters later (e.g. different endpoints and so on).

GabrielBianconi (changes requested):

gateway/src/endpoints/inference.rs
```diff
@@ -92,7 +120,7 @@ pub async fn inference_handler(
         http_client,
         clickhouse_connection_info,
     }): AppState,
-    StructuredJson(params): StructuredJson<Params>,
+    StructuredJson(params): StructuredJson<Params<'static>>,
```
GabrielBianconi, line 123: Skill issue on my side: how is this `'static`?

gateway/src/error.rs
```diff
@@ -400,6 +405,13 @@ impl std::fmt::Display for Error {
             Error::AzureServer { message } => {
                 write!(f, "Error from Azure server: {}", message)
             }
+            Error::BadCredentialsPreInference { provider_name } => {
+                write!(
+                    f,
+                    "Bad credentials at inference time for provider: {}. This should never happen.",
```
GabrielBianconi, line 411: For all `This should never happen` errors, append a `Please file a bug report: https://github.com/tensorzero/tensorzero/issues/new` to the message

gateway/src/inference/providers/aws_bedrock.rs
```diff
@@ -234,11 +240,22 @@ impl InferenceProvider for AWSBedrockProvider {
     }
 }
 
+lazy_static! {
+    static ref EMPTY_SECRET: SecretString = SecretString::from(String::new());
+}
```
GabrielBianconi, line 245: ?

gateway/src/inference/providers/fireworks.rs
```diff
@@ -149,6 +168,25 @@ impl HasCredentials for FireworksProvider {
     fn has_credentials(&self) -> bool {
         self.api_key.is_some()
     }
+    fn get_credentials<'a>(
```
GabrielBianconi, line 171: \n

GabrielBianconi (changes requested):
Claude-generated summary of Slack discussion:

We should consider implementing a `dynamic_credentials` flag for providers to handle cases where API keys are provided dynamically (e.g., not through environment variables). When this flag is set in `tensorzero.toml` (in a provider block):

1. Skip the initial API key validation check on startup
2. Ignore any environment variables / credential files for that provider to prevent conflicts
3. Ensure the provider uses either dynamic credentials or static (env var) credentials, not both

This would solve the issue of errors on startup when using dynamic credentials and provide a clear way to configure providers for different authentication methods.

gateway/src/inference/providers/provider_trait.rs
```diff
@@ -1,26 +1,34 @@
...
 use crate::inference::types::ProviderInferenceResponseStream;
+use crate::model::ProviderCredentials;
 use futures::Future;
 use reqwest::Client;
 
 pub trait HasCredentials {
     fn has_credentials(&self) -> bool;
+    fn get_credentials<'a>(
+        &'a self,
+        api_keys: &'a InferenceCredentials,
```
GabrielBianconi, line 15: Rename `api_keys` to `credentials` everywhere in the trait.

GabrielBianconi (changes requested):
Add E2E tests: for every provider...

* For a function with dynamic credential, not providing fails it...
* For the same function, providing the credentials makes it work.

gateway/src/endpoints/inference.rs
virajmehta: it's static because they're always owned here -> each of the API keys is a Cow <'a, SecretString> but it is only a reference if you grab it from env at runtime and it is static.

GabrielBianconi (approved):

PR #368, opened 2024-09-30
Fix openai json mode for non-strict

### LINKED ISSUE #141: Ensure JSON mode works even when prompt doesn't mention JSON
For providers (`openai`, `azure`) which return 400 on JSON mode without "json" in the prompt, add an instruction to use JSON to the prompt if it's not already there.

virajmehta:
Closes #141


gateway/src/inference/providers/fireworks.rs
```diff
@@ -362,6 +388,9 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_prepare_fireworks_messages() {}
```
GabrielBianconi, line 392: ?

gateway/src/inference/providers/openai.rs
```diff
@@ -382,23 +398,68 @@ pub(super) fn prepare_openai_tools<'a>(
...
+                            .map(|c| c.to_lowercase().contains("json"))
+                            .unwrap_or(false)
+                    }) || instructions.to_lowercase().contains("json")
+                    {
+                        OpenAIRequestMessage::System(OpenAISystemRequestMessage {
+                            content: Cow::Borrowed(instructions),
+                        })
+                    } else {
+                        let formatted_instructions =
+                            format!("{} Respond using JSON.", instructions);
```
GabrielBianconi, line 427: I think we should do `Respond using JSON.\n\n{instructions}`, in case there's something weird about the instructions.

Also you mix "system" and "instructions" but we don't use "instructions" elsewhere? `Some(system)` would be clearer.

GabrielBianconi (changes requested):

GabrielBianconi (approved):

PR #376, opened 2024-10-08
Improve client

### LINKED ISSUE #375: TensorZero python classes can't be passed back into `inference` for multistep flows
here's the sort of thing I would like to be able to do:

```python
async def solve_beerqa(client: AsyncTensorZeroGateway, question: str, query_budget: int = 3) -> str:
    """
    Solve a BeerQA question using Wikipedia.
    """
    messages = [{"role": "user", "content": question}]
    for queries_remaining in range(query_budget, 0, -1):
        response = await client.inference(function_name="beerqa_solver", input={
            "system": {"queries_remaining": queries_remaining},
            "messages": messages
        })
        # the following line puts a tensorzero Text and / or ToolCall object into the message history
        # these should be handled nicely by the client so that 
        messages.append({"role": "user", "content": response.content})
        for block in response.content:
            if block.type == "tool_call":
                if block.name == "submit_answer":
                    return block.arguments["answer"]
                result = await execute_tool_call(block)
                print(result)
```

virajmehta:
Will close #375

### DIFF
--- a/clients/python/src/tensorzero/__init__.py
+++ b/clients/python/src/tensorzero/__init__.py
@@ -11,6 +11,7 @@ from .types import (
     TextChunk,
     ToolCall,
     ToolCallChunk,
+    ToolResult,
     Usage,
 )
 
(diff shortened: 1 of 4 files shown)


GabrielBianconi (approved):

GabrielBianconi:
@ellipsis review this [testing richer PR description]

PR #385, opened 2024-10-10
Integrated Google AI Studio Gemini provider

### LINKED ISSUE #159: Integrate with the Gemini API outside GCP
The Gemini API documented [here](https://ai.google.dev/gemini-api/docs) seems better supported (in terms of model types) and documented. It also allows you to call with the typical API key instead of a JWT. We should probably eventually support both but I think we should eventually probably prefer this one.

virajmehta:
Closes #159


GabrielBianconi:
Add to `.example.env`

GabrielBianconi (changes requested):
See above. Check for every mention of a random credential like Together or something.

GabrielBianconi (approved):

PR #392, opened 2024-10-13
Implement prefilling for JSON mode for Anthropic models

virajmehta:


gateway/src/inference/providers/anthropic.rs
```diff
@@ -518,6 +539,62 @@ fn prepare_messages(messages: Vec<AnthropicMessage>) -> Result<Vec<AnthropicMess
...
+            id: chunk.inference_id.to_string(),
+        })];
+    } else if chunk.content.len() == 1 {
+        if let ContentBlockChunk::Text(TextChunk { text, .. }) = &chunk.content[0] {
+            // Add a "{" to the beginning of the text
+            chunk.content = vec![ContentBlockChunk::Text(TextChunk {
+                text: format!("{{{}", text.trim()),
+                id: chunk.inference_id.to_string(),
+            })];
+        }
```
GabrielBianconi, line 593: else with error handling?

gateway/src/inference/providers/anthropic.rs
```diff
@@ -518,6 +539,62 @@ fn prepare_messages(messages: Vec<AnthropicMessage>) -> Result<Vec<AnthropicMess
...
+    if chunk.content.is_empty() {
+        chunk.content = vec![ContentBlockChunk::Text(TextChunk {
+            text: "{".to_string(),
+            id: chunk.inference_id.to_string(),
+        })];
+    } else if chunk.content.len() == 1 {
+        if let ContentBlockChunk::Text(TextChunk { text, .. }) = &chunk.content[0] {
+            // Add a "{" to the beginning of the text
+            chunk.content = vec![ContentBlockChunk::Text(TextChunk {
+                text: format!("{{{}", text.trim()),
```
GabrielBianconi, line 590: You can't trim chunks. Imagine chunks `{"msg": "Hello "` + `world ".}`

Left-trim only. (Above should be ok but might as well do it right.)

GabrielBianconi (changes requested):
Repeat for every provider

gateway/src/inference/providers/anthropic.rs
GabrielBianconi: Add this test case

GabrielBianconi (approved):

PR #394, opened 2024-10-14
Implemented migrations for tagged feedback

virajmehta:


GabrielBianconi (changes requested):

gateway/src/clickhouse_migration_manager/migrations/migration_0003.rs
```diff
@@ -94,100 +94,169 @@ impl<'a> Migration for Migration0003<'a> {
...
             AS
                 SELECT
                     'comment' as metric_name,
-                    entry.1 as tag_key,
-                    entry.2 as tag_value,
+                    entry.1 as key,
+                    entry.2 as value,
                     id as feedback_id
                 FROM CommentFeedback
                 ARRAY JOIN arrayZip(mapKeys(tags), mapValues(tags)) AS entry
```
GabrielBianconi, line 232: Are you 100% sure this works? Where is this snippet from? I couldn't find in CH docs.

Failure case: `mapKeys` and `mapValues` don't use the same order for some reason.

Why not retrieve all the keys, then `k as key`, `map[k] as value` or similar?

GabrielBianconi (changes requested):

gateway/src/clickhouse_migration_manager/migrations/migration_0003.rs
virajmehta: it's from iteratively playing with ClickHouse -- I can change it to this as a more idiot-proof solution but I am entirely confident this does the job. WIll fix.

GabrielBianconi (approved):

PR #397, opened 2024-10-16
Add a production deployment example and also CI tests that check that it works

virajmehta:


GabrielBianconi (changes requested):
This is pulling `tensorzero/gateway` from Dockerhub, which is not the version in the PR.

We need to build the container in the GHA. (At that point, might as well run the E2E test inside it, so we don't build both inside and outside.)

.github/workflows/merge-queue.yml
```diff
@@ -51,11 +50,15 @@ jobs:
...
+      - name: Set up CLICKHOUSE_URL for E2E tests
+        run: |
+          echo "CLICKHOUSE_URL=http://localhost:8123/tensorzero" >> $GITHUB_ENV
+
       - name: Launch ClickHouse container for E2E tests
         run: docker compose -f gateway/tests/e2e/docker-compose.yml up -d --wait
 
       - name: Launch the gateway for E2E tests
-        run: cargo run-e2e &
+        run: cargo run-e2e & echo $! > .gateway.pid
```
GabrielBianconi, line 61: Clean

GabrielBianconi (approved):

PR #407, opened 2024-10-26
Add a recipe for fine-tuning on demonstrations using Fireworks

virajmehta:


GabrielBianconi (approved):
See my change. Updated FT job polling method so that you have the job ID in case something fails (you can resume).

virajmehta:
that tracks for me lgtm

PR #426, opened 2024-11-04
added uuid7 to client code

### LINKED ISSUE #420: Consider including a UUID7 implementation in the `tensorzero` Python package
Many users first try using `pip install uuid7` which is broken and abandoned (see https://github.com/stevesimmons/uuid7/issues/1). When using this library, they'll get `Invalid Episode ID: Timestamp is in the future` because the library is not handling the timestamp correctly.

There are good choices like https://github.com/aminalaee/uuid-utils but that can be confusing.

Maybe we should include a UUID7 implementation (e.g. adding a dependecy or vendoring a correct implementation) in our Python package?

`from tensorzero.util import uuid7`?

virajmehta:
Closes #420 

Also removes a dependency on the incorrect implementation of uuidv7 (from `pip install uuid7`).


GabrielBianconi (approved):

GabrielBianconi (changes requested):
Can you also update the bad `uuid7` library from elsewhere in the repo? And make sure every Python-based UUID generation now uses this (instead of either external lib)?

GabrielBianconi (approved):

.github/workflows/merge-queue.yml
```diff
@@ -2,6 +2,7 @@ name: Merge Queue Checks
 
 on:
   merge_group:
+  push:
```
GabrielBianconi, line 5: This needs to go before merging

GabrielBianconi: (But fix it first)

PR #429, opened 2024-11-05
Allow text blocks alongside tool calls in provider tests

### LINKED ISSUE #428: Tests are failing more than is needed when models return text + tool call
We should assert that there is 1 tool call and that the tool call is as desired but not assert that there is no text also

virajmehta:
Closes #428


GabrielBianconi (approved):
`panic!("Unreachable")`

PR #440, opened 2024-11-09
OpenAI fine tuning for boolean metrics

virajmehta:
This is not necessarily mergeable but it shows how to actually do the fine-tune in TypeScript. We will likely want to add some tests and also determine how to build a broader web application to enable this more easily. 

But, I think it's worth reviewing at this point and running to make sure you can. Let's sync up on how to incorporate this into the plan. I can now easily fill out the rest of feature-parity for fine-tuning here.


dashboard/utils/fine_tuning/rendering.ts
```diff
@@ -0,0 +1,32 @@
...
+) {
+  if (role !== "user" && role !== "assistant") {
+    throw new Error('Role must be either "user" or "assistant"');
+  }
+  if (content.length !== 1) {
+    throw new Error("Content must contain exactly one block");
+  }
+  if (content[0].type !== "text") {
+    throw new Error("Content must be a text block");
+  }
```
GabrielBianconi, line 27: This will break with tools. Either implement or log somewhere as a TODO.

dashboard/utils/fine_tuning/rendering.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @throws {Error} If role is invalid, content length is not 1, or content type is not text
+ */
+export function render_message(
+  env: JsExposedEnv,
+  role: Role,
+  content: InputMessageContent[],
+) {
+  if (role !== "user" && role !== "assistant") {
+    throw new Error('Role must be either "user" or "assistant"');
+  }
```
GabrielBianconi, line 21: Not needed since role is an enum?

dashboard/utils/fine_tuning/rendering.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (content.length !== 1) {
+    throw new Error("Content must contain exactly one block");
+  }
+  if (content[0].type !== "text") {
+    throw new Error("Content must be a text block");
+  }
+  if (typeof content[0].value === "string") {
+    return content[0].value;
+  }
```
GabrielBianconi, line 30: Same, for schemas?

dashboard/utils/fine_tuning/openai.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const client = new OpenAI({
+  apiKey: process.env.OPENAI_API_KEY,
+});
+
+export function sample_to_openai_messages(
+  sample: ParsedInferenceRow,
+  env: JsExposedEnv,
+) {
```
GabrielBianconi, line 27: `tensozero_inference_to_openai_messages`

dashboard/utils/fine_tuning/openai.ts
```diff
@@ -0,0 +1,116 @@
...
+    });
+  }
+  for (const message of sample.input.messages) {
+    const rendered_message = render_message(env, message.role, message.content);
+    messages.push({
+      role: message.role,
+      content: rendered_message,
+    });
+  }
+  // Add type check at the beginning
```
GabrielBianconi, line 54: ?

dashboard/utils/fine_tuning/openai.ts
```diff
@@ -0,0 +1,116 @@
...
+  if (isChatInference) {
+    const output = sample.output as ContentBlockOutput[];
+    if (output.length !== 1) {
+      throw new Error("Chat inference must have exactly one message");
+    }
+    if (output[0].type !== "text") {
+      throw new Error("Chat inference must have a text message as output");
+    }
+    messages.push({ role: "assistant", content: output[0].text });
+  } else {
```
GabrielBianconi, line 65: Don't else here. if/elseif/error. in case we add more function types.

dashboard/utils/fine_tuning/openai.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  for (const message of sample.input.messages) {
+    const rendered_message = render_message(env, message.role, message.content);
+    messages.push({
+      role: message.role,
+      content: rendered_message,
+    });
+  }
+  // Add type check at the beginning
+  const isChatInference = Array.isArray(sample.output);
```
GabrielBianconi, line 55: This is not the right way to do this

dashboard/utils/fine_tuning/openai.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export async function upload_examples_to_openai(samples: OpenAIMessage[][]) {
+  // Convert samples to JSONL format
+  const jsonl = samples
+    .map((messages) => JSON.stringify({ messages }))
+    .join("\n");
+
+  // Write to temporary file
+  const tempFile = path.join(process.cwd(), "temp_training_data.jsonl");
+  await fs.writeFile(tempFile, jsonl);
```
GabrielBianconi, line 80: This is not safe (tabs overwrite, etc.). Write to `/tmp` using a random identifier, then delete as you do.

dashboard/utils/fine_tuning/openai.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  try {
+    const file = await client.files.create({
+      file: createReadStream(tempFile),
+      purpose: "fine-tune",
+    });
+
+    return file.id;
+  } finally {
```
GabrielBianconi, line 89: This doesn't run on success I think since you're returning...?

dashboard/utils/fine_tuning/rendering.ts
virajmehta: implemented elsewhere since it's openai-specific. Rendering only makes sense for text messages in our data model.

dashboard/utils/fine_tuning/rendering.ts
virajmehta: still getting used to TypeScript semantics

virajmehta: removed

dashboard/utils/fine_tuning/openai.ts
virajmehta: I think it does run according to multiple LLMs?

virajmehta: confirmed from reading docs

dashboard/utils/fine_tuning/openai.ts
virajmehta: I think the right way to do it is actually pass the function type in. We can add this as we actually wire the whole thing up.

dashboard/utils/fine_tuning/openai.ts
GabrielBianconi: wow

dashboard/utils/fine_tuning/openai.ts
GabrielBianconi: ok

dashboard/utils/fine_tuning/rendering.ts
GabrielBianconi: nvm

GabrielBianconi:
Final PR #520

PR #444, opened 2024-11-12
Implement an OpenAI-compatible TensorZero endpoint

### LINKED ISSUE #423: Proposal: Implement an OpenAI-compatible endpoint for the gateway
## Motivation

LLM users and existing tools most commonly use the OpenAI API. TensorZero currently has an API that maps onto our internal representations, but we should also offer an OpenAI-compatible endpoint that still allows us to provide as much of the TensorZero functionality as possible.

### Background
As shown below, the OpenAI client allows for message content of the form of a list containing a dict, so we can support `Input` as in TensorZero with some type-casting. We can pass additional data to the server using the headers and parse them server-side. 
```python
    response = await client.chat.completions.create(
    messages=[
        {
            "role": "user",
            "content": [{
                "x": "y",
                "a": "b",
            }],
        }
    ],
    model="tensorzero",
    extra_headers={
        "episode_id": "...",
        "function_name": "write_haiku",
    },
)

## Body on server side of this request
# {"messages": [{"role": "user", "content": [{"x": "y", "a": "b"}]}], "model": "tensorzero"}
# Also should include the headers passed in `extra_headers`
```

### Proposed Interface input
We need a layer mapping the OpenAI-compatible interface on the server-side to our typical Input.

Here's the set of request parameters our current inference endpoint expects along with where we expect to take them from in this proposal:

```rust
/// The expected payload is a JSON object with the following fields:
#[derive(Debug, Deserialize)]
#[serde(deny_unknown_fields)]
pub struct Params<'a> {
    // the function name
    function_name: String, // headers with parameter function_name (required) 
    // the episode ID (if not provided, it'll be set to inference_id)
    // NOTE: DO NOT GENERATE EPISODE IDS MANUALLY. THE API WILL DO THAT FOR YOU.
    episode_id: Option<Uuid>, // headers (optional)
    // the input for the inference
    input: Input,  // constructed from messages (see below)
    // default False
    stream: Option<bool>, // from OpenAI request parameter
    // Inference-time overrides for variant types (use with caution)
    #[serde(default)]
    params: InferenceParams,  // from OpenAI request parameters (chat completions only)
    // if the client would like to pin a specific variant to be used
    // NOTE: YOU SHOULD TYPICALLY LET THE API SELECT A VARIANT FOR YOU (I.E. IGNORE THIS FIELD).
    //       ONLY PIN A VARIANT FOR SPECIAL USE CASES (E.G. TESTING / DEBUGGING VARIANTS).
    variant_name: Option<String>,  // from headers (optional)
    // if true, the inference will not be stored
    dryrun: Option<bool>,  // from headers (optional)
    // dynamic information about tool calling. Don't directly include `dynamic_tool_params` in `Params`.
    #[serde(flatten)]
    dynamic_tool_params: DynamicToolParams, 
    // `dynamic_tool_params` includes the following fields, passed at the top level of `Params`:
    // If provided, the inference will only use the specified tools (a subset of the function's tools)
    // allowed_tools: Option<Vec<String>>, // Assumed to be empty
    // If provided, the inference will use the specified tools in addition to the function's tools
    // additional_tools: Option<Vec<Tool>>, // from tools parameter in OpenAI API
    // If provided, the inference will use the specified tool choice
    // tool_choice: Option<ToolChoice>, // from tool_choice parameter in OpenAI API
    // If true, the inference will use parallel tool calls
    // parallel_tool_calls: Option<bool>, // from parallel_tool_calls parameter in OpenAI API
    // If provided for a JSON inference, the inference will use the specified output schema instead of the
    // configured one. We only lazily validate this schema.
    output_schema: Option<Value>, // as used for structured json mode in OpenAI
    #[serde(default)]
    credentials: InferenceCredentials<'a>,  // optional request header
}
```

### Proposed Interface Output mapping
I'll comment what I think we should put in each field and how we'll get it.
```python
{
  "id": "chatcmpl-123456", # Inference ID
  "object": "chat.completion", # hardcoded
  "created": 1728933352, # timestamp
  "model": "gpt-4o-2024-08-06", # variant_name
  "choices": [
    {
      "index": 0, # hardcoded
      "message": {
        "role": "assistant",
        "content": "Hi there! How can I assist you today?", # Text content blocks
        "refusal": null,
        "tool_calls": [{"id": "tool_call_id", "type": "function", "function": {"name": str, "arguments": str}] # Array of Tool calls if applicable
      },
      "logprobs": null, # Always null
      "finish_reason": "stop" # alwasy
    }
  ],
  "usage": {
    "prompt_tokens": 19, # As needed
    "completion_tokens": 10,  # As needed
    "total_tokens": 29, # As needed
    "prompt_tokens_details": {
      "cached_tokens": 0
    },
    "completion_tokens_details": {
      "reasoning_tokens": 0
    }
  },
  "system_fingerprint": "fp_6b68a8204b"
}
```

Streaming output: 
```
{
  "id": "chatcmpl-8sn4n2n3j2n3j2n3",  # Inference id
  "object": "chat.completion.chunk", # hardcoded
  "created": 1709548800, # timestamp
  "model": "claude-3-sonnet-20240229", # function_name
  "system_fingerprint": "fp_44795b6733", # idk
  "service_tier": "standard", # idk
  "choices": [
    {
      "index": 0, # hardcoded
      "delta": {
        "content": "Here's how you can calculate", # text content if present
        "role": "assistant", # hardcoded
        "tool_calls": [
          {
            "index": 0, # hardcoded
            "id": "call_4k2j34k2j34k2j", # from ToolCall content block
            "type": "function",
            "function": {
              "name": "calculate_average", # from ToolCall content block
              "arguments": "{\"numbers\": [1, 2, 3, 4, 5]}" # from ToolCall content block
            }
          }
        ],
        "refusal": null # hardcoded
      },
 
        ]
      },
      "finish_reason":null # hardcoded 
    }
  ],
  "usage": {
    "prompt_tokens": 156,
    "completion_tokens": 423,
    "total_tokens": 579
  }
}
```

### Testing
First, write tests with OpenAI client in Python to cover every case in `/gateway/tests/e2e/inference.rs` that makes sense using dummy models and assertions. Here are the cases we should tackle:

 * simple inference request
 * simple streaming inference request
 * inference params request
 * tool use additional tool
 * tool use tool choice (each kind)
 * streaming tool use
 * multiturn tool use
 * parallel tool use
 * parallel streaming tool use
 * json function
 * dynamic json function
 * streaming json function

virajmehta:
Closes #423


gateway/src/endpoints/openai_compatible.rs
```diff
@@ -0,0 +1,1020 @@
...
+            function_name,
+            episode_id,
+            input,
+            stream: openai_compatible_params.stream,
+            params: inference_params,
+            variant_name,
+            dryrun,
+            dynamic_tool_params,
+            output_schema,
+            credentials: InferenceCredentials::default(),
```
GabrielBianconi, line 324: Why not support this, at least for OpenAI? Or document we don't for now.

gateway/src/endpoints/openai_compatible.rs
```diff
@@ -0,0 +1,1020 @@
...
+        let mut tool_call_id_to_name = HashMap::new();
+        for message in openai_compatible_messages {
+            match message {
+                OpenAICompatibleMessage::System(msg) => {
+                    if system.is_some() {
+                        return Err(Error::InvalidOpenAICompatibleRequest {
+                            message: "At most one system message is allowed".to_string(),
+                        });
+                    }
+                    system = Some(convert_openai_message_content(msg.content)?);
```
GabrielBianconi, line 345: This should explicitly check it's the first, or fail.

GabrielBianconi (changes requested):

gateway/src/endpoints/openai_compatible.rs
virajmehta: I find it unintuitive to only support one kind of credential and it is not ergonomic to put more complicated JSON objects in headers. I will document that we don't for now

gateway/src/endpoints/openai_compatible.rs
virajmehta: I dont' see why we can't do best-effort with that

GabrielBianconi: we should not reorder messages without the user knowing

gateway/src/endpoints/openai_compatible.rs
GabrielBianconi: we can snooze

GabrielBianconi (changes requested):
Per discussion, let's use the `model` field to specify the `function_name` as `tensorzero::FUNCTION_NAME`.

PR #459, opened 2024-11-13
added instrumentation to annotate error logging for model failures and such

virajmehta:


gateway/src/clickhouse.rs
```diff
@@ -143,6 +144,7 @@ impl ClickHouseConnectionInfo {
         }
     }
 
+    #[instrument(skip(self))]
     pub async fn run_query(&self, query: String) -> Result<String, Error> {
```
GabrielBianconi, line 148: We shouldn't log this function as it could contain sensitive data in the query. Either skip_all or remove.

GabrielBianconi (changes requested):

gateway/src/function.rs
```diff
@@ -124,6 +125,13 @@ impl FunctionConfig {
         }
     }
 
+    #[instrument(skip(
+        content_blocks,
+        usage,
+        model_inference_results,
+        inference_config,
+        inference_params
+    ))]
```
GabrielBianconi, line 134: You're leaking stuff here (e.g. `self`). `skip_all` + explicitly select what you want.

I haven't tested but think the following should work.

```
#[instrument(skip_all, fields(inference_id))]
```

GabrielBianconi (changes requested):

gateway/src/function.rs
virajmehta: that works!

PR #462, opened 2024-11-14
Change default startup behavior to not require ClickHouse

virajmehta:


gateway/src/main.rs
```diff
@@ -19,16 +20,15 @@ static GLOBAL: MiMalloc = MiMalloc;
 async fn main() {
     // Set up logs and metrics
     observability::setup_logs();
-    let metrics_handle = observability::setup_metrics().expect_pretty("Failed to set up metrics");
+    let metrics_handle = observability::setup_metrics().ok_or_log();
```
GabrielBianconi, line 23: This should still fail i think, nothing to do with ClickHouse. (And generally shouldn't even fail.)

gateway/src/main.rs
```diff
@@ -46,11 +46,14 @@ async fn main() {
...
-        .route(
+        .with_state(app_state);
+
+    if let Some(metrics_handle) = metrics_handle {
+        router = router.route(
             "/metrics",
             get(move || std::future::ready(metrics_handle.render())),
         )
-        .with_state(app_state);
+    }
```
GabrielBianconi, line 56: Metrics shouldn't fail because of clickhouse, no need?

gateway/src/gateway_util.rs
```diff
@@ -17,24 +17,36 @@ pub struct AppStateData {
...
+        let clickhouse_url = std::env::var("CLICKHOUSE_URL")
+            .map_err(|_| Error::AppState {
                 message: "Missing environment variable CLICKHOUSE_URL".to_string(),
-            })?;
-
-            ClickHouseConnectionInfo::new(&clickhouse_url)?
-        };
+            })
+            .ok_or_log()
+            .unwrap_or_else(|| "".to_string());
```
GabrielBianconi, line 26: It should not log if variable is missing but observability is disabled

GabrielBianconi: Move this into `setup_clickhouse` in the `else` branch.

GabrielBianconi (changes requested):

gateway/src/main.rs
virajmehta: I was trying to be as conservative as possible about keeping the gateway up but I guess it's OK

virajmehta: After investigating, this only fails if you've already exported the metrics. OK to keep an expect

nikcaryo-super (approved):

PR #473, opened 2024-11-19
added instrumentation & tracing for top-level handlers, enforced nice…

virajmehta:
… minijinja errors


GabrielBianconi (approved):
small change

PR #489, opened 2024-11-26
Handle streaming failures correctly

GabrielBianconi:


clients/python/src/tensorzero/client.py
```diff
@@ -493,23 +487,30 @@ async def __aexit__(
...
-                data = line[6:].strip()
-                if data == "[DONE]":
-                    break
-                try:
-                    parsed_data: Dict[str, Any] = json.loads(data)
-                    yield parse_inference_chunk(parsed_data)
-                except json.JSONDecodeError:
-                    self.logger.error(f"Failed to parse SSE data: {data}")
-        await response.aclose()
+        async with self.client.stream("POST", url, json=data) as response:
```
virajmehta, line 500: can you add documentation of the weird httpx behavior here?

clients/python/src/tensorzero/client.py
```diff
@@ -327,26 +321,33 @@ def __exit__(
...
-                    self.logger.error(f"Failed to parse SSE data: {data}")
-        response.close()
+        with self.client.stream("POST", url, json=data) as response:
+            try:
+                response.raise_for_status()
+            except httpx.HTTPStatusError as e:
+                response.read()
+                raise TensorZeroError(response) from e
+
+            for line in response.iter_lines():
```
virajmehta, line 341: can you add a comment documenting weird httpx streaming behavior here?

clients/python/tests/test_client.py
```diff
@@ -85,14 +84,18 @@ async def test_async_inference_streaming(async_client):
...
+        previous_chunk_timestamp = time()
         chunks.append(chunk)
-        if first_chunk_duration is None:
-            first_chunk_duration = time() - start_time
-    last_chunk_duration = time() - start_time - first_chunk_duration
-    assert last_chunk_duration > first_chunk_duration + 0.1
+
+    assert last_chunk_duration > 0.01
+
     expected_text = [
```
virajmehta, line 99: am I correct in that we were already testing this then?

clients/python/tests/test_client.py
```diff
@@ -546,17 +559,42 @@ def test_sync_inference_streaming(sync_client):
...
             input={
                 "system": {"assistant_name": "Alfred Pennyworth"},
                 "messages": [{"role": "user", "content": "Hello"}],
             },
             stream=True,
         )
+
+        for chunk in stream:
+            pass
+
```
virajmehta, line 573: comment why

virajmehta (approved):
small questions and doc fixes otherwise LGTM

clients/python/tests/test_client.py
GabrielBianconi: yeah, but somewhat incorrectly (we were checking the duration of all chunks after the first combined)

Issue #515, opened 2024-12-11
Migrate from `mypy` to `pyright`


GabrielBianconi:
* Migrate from `mypy` to `pyright`
* Run `pyright` on `pre-commit`
* Run `pyright` on PR commit checks (`general.yml`)

amerinor01:
Hi, I'm interested to take this one. @GabrielBianconi, can I take it?

GabrielBianconi:
@amerinor01 Sounds good, thanks!

amerinor01:
Hi, I'm working on but I have some question. Pyright have some flag call `strict` to make it more pedantic about the errors and warnings. 

By setting the flag to the file `clients\python\client.py` you have more errors than previously.

```
/home/a/tensorzero/clients/python/src/tensorzero/client.py
  /home/a/tensorzero/clients/python/src/tensorzero/client.py:91:24 - error: Type of "item" is unknown (reportUnknownVariableType)
  /home/a/tensorzero/clients/python/src/tensorzero/client.py:91:42 - error: Argument type is partially unknown
    Argument corresponds to parameter "iterable" in function "__new__"
    Argument type is "list[Unknown]" (reportUnknownArgumentType)
  /home/a/tensorzero/clients/python/src/tensorzero/client.py:92:32 - error: Argument type is unknown
    Argument corresponds to parameter "obj" in function "hasattr" (reportUnknownArgumentType)
  /home/a/tensorzero/clients/python/src/tensorzero/client.py:93:49 - error: Type of "to_dict" is unknown (reportUnknownMemberType)
4 errors, 0 warnings, 0 informations 
```

@GabrielBianconi  you think I should fix it pre pull, or make the pipeline to fail?

GabrielBianconi:
Hi @amerinor01 - try to fix if it's simple, otherwise open the pull request and we can fix. Thanks!

GabrielBianconi:
Thanks @amerinor01!

PR #551, opened 2024-12-19
#515 fixed mypy to pyright

gautamraj8044:
> [!IMPORTANT]
> Replaces `mypy` with `pyright` for type checking in `merge-queue.yml`.
> 
>   - **Type Checking**:
>     - Replaces `mypy` with `pyright` in `merge-queue.yml`.
>     - Updates installation command from `uv pip install mypy` to `uv pip install pyright`.
>     - Updates execution command from `uv run mypy python --strict` to `uv run pyright python --strict`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 97e40f1a1f4517e4e495e7bfc51d5ed0d9656b9e. It will automatically update as commits are pushed.</sup>


GabrielBianconi (changes requested):
Hi - I believe the proposed changes don't work. `pyright` doesn't have a `--strict` flag for example. We also need to update `.pre-commit-config.yaml`. Please update this file and try running `pre-commit` to ensure it works.

Issue #383, opened 2024-10-09
implement Anthropic `disable_parallel_tools` flag


virajmehta:

hk1997:
Hi,
I would like to pick this feature.
From what I understand we are trying to support the following functionality: https://docs.anthropic.com/en/docs/build-with-claude/tool-use#disabling-parallel-tool-use

The issue calls for changes in gateway/src/tool.rs & anthropic provider if I'm understanding this correctly.

GabrielBianconi:
@hk1997 yes!

hk1997:
@GabrielBianconi I reckon we can mark this issue as closed since https://github.com/tensorzero/tensorzero/pull/557 is merged.

GabrielBianconi:
Thank you for the contribution @hk1997!

PR #557, opened 2024-12-20
Add support for disable_parallel_tool_use flag in

hk1997:
implement Anthropic disable_parallel_tools flag

Issue: https://github.com/tensorzero/tensorzero/issues/383

----

> [!IMPORTANT]
> Add `disable_parallel_tool_use` flag to `AnthropicRequestBody` and update tests in `anthropic.rs`.
> 
>   - **Behavior**:
>     - Add `disable_parallel_tool_use` flag to `AnthropicRequestBody` in `anthropic.rs`.
>     - Set `disable_parallel_tool_use` based on `tool_config.parallel_tool_calls` in `AnthropicRequestBody::new()`.
>   - **Tests**:
>     - Update tests in `tests` module to include `disable_parallel_tool_use` flag checks.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for aee72890496fe137a6d69e44c4c587c28180cc2b. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gateway/src/inference/providers/anthropic.rs
+++ b/gateway/src/inference/providers/anthropic.rs
@@ -23,7 +23,7 @@ use crate::inference::types::{
     ProviderInferenceResponseStream, RequestMessage, TextChunk, Usage,
 };
 use crate::model::CredentialLocation;
-use crate::tool::{ToolCall, ToolCallChunk, ToolChoice, ToolConfig};
+use crate::tool::{ToolCall, ToolCallChunk, ToolCallConfig, ToolChoice, ToolConfig};
 
 lazy_static! {
     static ref ANTHROPIC_BASE_URL: Url = {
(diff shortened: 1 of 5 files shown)


gateway/src/inference/providers/anthropic.rs
```diff
@@ -1721,6 +1732,7 @@ mod tests {
             top_p: Some(0.9),
             tool_choice: None,
             tools: None,
+            disable_parallel_tool_use: None,
```
virajmehta, line 1735: can you add test cases in here where the ToolConfig is set with parallel_tool_calls=true and parallel_tool_calls=false?

virajmehta (changes requested):
Generally looks good but we should have a unit test case covering this behavior. Thanks!

gateway/src/inference/providers/anthropic.rs
hk1997: Good call, while the tests for true case were already there, we were missing on false behavior.
Added an explicit test to test both the behaviors.

virajmehta: yep, will do one pass of reviews and then merge! great work

virajmehta (approved):
LGTM

gateway/src/inference/providers/anthropic.rs
```diff
@@ -448,6 +448,8 @@ struct AnthropicRequestBody<'a> {
     tool_choice: Option<AnthropicToolChoice<'a>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     tools: Option<Vec<AnthropicTool<'a>>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    disable_parallel_tool_use: Option<bool>,
 }
```
virajmehta, line 453: hey @hk1997 sorry I didn't catch this earlier but the tests did: `disable_parallel_tool_use` actually goes in the `AnthropicToolChoice` struct as a struct field, so 

```rust
/// We can instruct Anthropic to use a particular tool,
/// any tool (but to use one), or to use a tool if needed.
#[derive(Clone, Debug, PartialEq, Serialize)]
#[serde(tag = "type")]
#[serde(rename_all = "snake_case")]
enum AnthropicToolChoice<'a> {
    Auto { disable_parallel_tool_use: Option<bool> },
    Any { disable_parallel_tool_use: Option<bool> },
    Tool { name: &'a str, disable_parallel_tool_use: Option<bool> },
}
```

so we'll need to move this into there and then change the TryFrom to take a reference to the whole ToolConfig. 
Do you mind making that change?

virajmehta (changes requested):

gateway/src/inference/providers/anthropic.rs
hk1997: Glad to have good integration tests in place!

Yes I can make the change but can you help clarify the following:

Isn't the purpose of disable_parallel_tool_use is to enforce a sequential use of provided tools, and is not a localized behavior of a particular tool?

If I'm not wrong the curl request would look as follows:
```
curl https://api.anthropic.com/v1/messages \
  -H "x-api-key: YOUR_API_KEY" \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -d '{
    "model": "claude-3-opus-20240229",
    "messages": [
      {
        "role": "user",
        "content": "Find information about the Anthropics API using the web tool."
      }
    ],
    "tools": [
      {"name": "web"},
      {"name": "python"}
    ],
    "max_tokens": 500,
    "disable_parallel_tool_use": true
}'
```

Am I missing something?

virajmehta: I don't disagree with this high-level reasoning.

However, 2 points:
*  the [api docs](https://docs.anthropic.com/en/api/messages#body-tool-choice) clearly show that the field goes in the tool choice block
* tool choice also has this property that this applies to the entire request. 
 
So doing a `TryFrom` on `&ToolCallConfig`  means you'll grab the tool choice from the `tool_choice` member and the `parallel_tool_choice` also and from there construct the enum above.
Does this make sense?

hk1997: Ahh yes, I missed the tool choice intent in the documentation.
Thank you for the clarification.

Yes it makes sense, I'll make the relevant changes.

virajmehta: thank you!

virajmehta:
@hk1997 this was great work, I added the e2e tests that this unlocked and ran `rustfmt`

virajmehta (approved):

GabrielBianconi (approved):

Issue #532, opened 2024-12-15
Duplicate logs are created for providers like fireworks using a mapping from openai error


Kaboomtastic:
Several providers are using functions like `map_openai_to_*_error` to map openai error types to the provider's error type but this generates a duplicate entry in the log with a message attributing the error to OpenAI first. In this example I provided an invalid api key to a fireworks provider and you can see in the first message that it is an error with a fireworks variant but the error type is OpenAI client

```
{"timestamp":"2024-12-15T19:37:27.152133Z","level":"WARN","message":"Error from OpenAI client: {\"error\":\"unauthorized\"}","spans":[{"function_name":"haiku","variant_name":"None","name":"inference"},{"function_name":"haiku","variant_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer"},{"model_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer_model_request"}]}
{"timestamp":"2024-12-15T19:37:27.152226Z","level":"WARN","message":"Error from Fireworks client: {\"error\":\"unauthorized\"}","spans":[{"function_name":"haiku","variant_name":"None","name":"inference"},{"function_name":"haiku","variant_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer"},{"model_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer_model_request"}]}
```

virajmehta:
Good issue! This is caused by the pattern of logging on error construction we adopted a few weeks ago in order to make sure the spans were attached properly. I think the fix might be to just make the client and server errors for inference providers generic, remove all the error type mapping, and be sure to attach spans that include the model name and provider type so the information is present. 

I could imagine this actually reducing the amount of code in T0 by a bit which is a nice side effect. @GabrielBianconi wdyt?

virajmehta:
Yeah! I was gonna do it soon but you're welcome to it. Can review whenever you're ready.

virajmehta:
Closed by #572

PR #572, opened 2024-12-24
#532 improve error logging

Kaboomtastic:
I consolidated existing error types into the existing InferenceClient and a new InferenceServer error types. I made status_code optional on InferenceClient and set it where there was a reqwest::Error to get the code from but I can change the field to be required and use a default value when it is missing. I added one additional span listing the provider name and prevented duplicates from getting emitted

This is the new log
`{"timestamp":"2024-12-24T03:47:30.708369Z","level":"ERROR","message":"{\"error\":\"unauthorized\"}","spans":[{"function_name":"haiku","variant_name":"None","name":"inference"},{"function_name":"haiku","variant_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer"},{"model_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer_model_request"},{"provider_name":"fireworks","name":"infer"}]}`

This is what would previously get emitted
`{"timestamp":"2024-12-15T19:37:27.152133Z","level":"WARN","message":"Error from OpenAI client: {\"error\":\"unauthorized\"}","spans":[{"function_name":"haiku","variant_name":"None","name":"inference"},{"function_name":"haiku","variant_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer"},{"model_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer_model_request"}]}
{"timestamp":"2024-12-15T19:37:27.152226Z","level":"WARN","message":"Error from Fireworks client: {\"error\":\"unauthorized\"}","spans":[{"function_name":"haiku","variant_name":"None","name":"inference"},{"function_name":"haiku","variant_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer"},{"model_name":"accounts/fireworks/models/llama-v3p2-3b-instruct","name":"infer_model_request"}]}`

----

> [!IMPORTANT]
> Consolidate error types into `InferenceClient` and `InferenceServer`, make `status_code` optional, and improve logging with provider name spans.
> 
>   - **Error Handling**:
>     - Consolidate error types into `InferenceClient` and `InferenceServer` in `error.rs`.
>     - Make `status_code` optional in `InferenceClient`.
>   - **Logging**:
>     - Add span for provider names in `model.rs` and `variant/mod.rs`.
>     - Prevent duplicate log entries.
>   - **Provider Updates**:
>     - Update error handling in `anthropic.rs`, `aws_bedrock.rs`, and `azure.rs` to use new error types.
>     - Similar updates in `fireworks.rs`, `gcp_vertex_anthropic.rs`, and `gcp_vertex_gemini.rs`.
>     - Apply changes to `google_ai_studio_gemini.rs`, `hyperbolic.rs`, and `mistral.rs`.
>     - Update `openai.rs`, `together.rs`, `vllm.rs`, and `xai.rs` for consistency.
>   - **Miscellaneous**:
>     - Update tests in `chat_completion.rs` and `variant/mod.rs` to reflect error handling changes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for c789197d8d17b63cfb71c0cd4727f723f0915475. It will automatically update as commits are pushed.</sup>


gateway/src/error.rs
```diff
@@ -542,35 +400,12 @@ impl std::fmt::Display for ErrorDetails {
...
-            ErrorDetails::HyperbolicClient { message, .. } => {
-                write!(f, "Error from Hyperbolic client: {}", message)
-            }
-            ErrorDetails::HyperbolicServer { message } => {
-                write!(f, "Error from Hyperbolic server: {}", message)
-            }
             ErrorDetails::Inference { message } => write!(f, "{}", message),
-            ErrorDetails::InferenceClient { message } => write!(f, "{}", message),
+            ErrorDetails::InferenceClient { message, .. } => write!(f, "{}", message),
+            ErrorDetails::InferenceServer { message, .. } => write!(f, "{}", message),
```
GabrielBianconi, line 408: [for discussion - @Kaboomtastic no need to update yet, cc @virajmehta]

Maybe we should add a prefix to this error message, like the old `Error from XYZ client/server"?

Potentially add a field to the error struct that contains the provider, so we can include that in the error message?

```
Error from inference [client|server] (model.model_name.provider.provider_name): error message (status code: xxx)
```

The span has quite a bit of info but the error message returned to the client doesn't.

GabrielBianconi:
Thank you for your contribution! @Kaboomtastic 

Superficially it looks good. We'll review it carefully soon.

gateway/src/error.rs
virajmehta: I think this is reasonable but shouldn't block merging. I will open another issue as I review.

GabrielBianconi: IMO we should address it because otherwise the error message will be very mysterious on the client side. If I understand this correctly, the client will only receive messages like `{\"error\":\"unauthorized\"}` - unclear if this is a gateway issue or downstream issue.

virajmehta: @Kaboomtastic I actually agree with @GabrielBianconi here on second thought -- we should add a field to the InferenceClientError and InferenceServerError that contains the provider type "openai"|"anthropic"|... so that we can use this in the error message, i.e. 

```rust
 ErrorDetails::InferenceClient { message, provider_type, .. } => write!(f, "Error from {} client: {}", provider_type, message),
 ErrorDetails::InferenceServer { message, provider_type, .. } => write!(f, "Error from {} server: {}", provider_type, message),
```

this would help people understand what isn't working. Do you mind adding this?

virajmehta:
Reviewed the rest of the PR and it looks good to me. Please add the change I requested and we're good to merge I think.

virajmehta (changes requested):
Reviewed the rest of the PR and it looks good to me. Please add the change I requested re: the provider type in the error structs and we're good to merge I think.

gateway/src/error.rs
virajmehta: You don't have that information at the time the error is created. We would have to build the string at runtime and change the `provider_trait` interface to do so OR change the `ProviderConfig` struct so that it contains the string and we can get it with `self.config_path`.  This would require us to build this string at initialization time for each model provider. I think the latter is better and easier to test.  I think this is gonna be too involved for now but worth doing soon. Maybe we do the simple thing now as I suggested and then add this as an issue?

GabrielBianconi: Sounds good

Kaboomtastic: I can add the provider and model name to the error message. I don't know if it's in scope of this issue but I think there should be either some function or constant that is associated to each implementor of Provider that is used for provider names universally.

virajmehta: you mean for them to render nicely? agree that that would be better, but for now let's hardcode it since we want to do paths eventually. thank you!

james-backpack:
Should be ready for review. I added the provider to the error types and to the error logs. I'm leaving for vacation for a couple weeks in a few days so I won't have time to make any additional changes if they're needed

virajmehta (dismissed):
LGTM! Merry Christmas

GabrielBianconi (dismissed):

GabrielBianconi (approved):

PR #576, opened 2024-12-24
Adding validations for inference id in feedback.rs.

hk1997:
For Inference level feedbacks, the inference id should be valid.
Adding logic to test that the inference id exists in InferenceById table.

Updated e2e tests and unit tests for the validation.

Issue: https://github.com/tensorzero/tensorzero/issues/350

----

> [!IMPORTANT]
> Add validation for inference IDs in feedback handling to ensure they exist in the `InferenceById` table, with updated tests to verify this behavior.
> 
>   - **Validation**:
>     - Added `get_target_identifier()` in `feedback.rs` to validate inference IDs against `InferenceById` table.
>     - Updated `write_comment()`, `write_demonstration()`, `write_float()`, and `write_boolean()` to use `get_target_identifier()` for ID validation.
>   - **Tests**:
>     - Updated unit tests in `feedback.rs` to cover new validation logic.
>     - Enhanced e2e tests in `feedback.rs` to ensure inference ID validation works correctly, rejecting invalid IDs and accepting valid ones.
>   - **Misc**:
>     - Removed `get_function_name_from_inference_id()` as its functionality is now covered by `get_target_identifier()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for f20b356e98980c4d2df5c0422d0c9f312c38f084. It will automatically update as commits are pushed.</sup>


virajmehta:
@hk1997 let me know when this is ready. I put myself on pre-emptively

hk1997:
@virajmehta It's ready for review now.

Ran into a lot of un-necessary changes due to merge conflict :( leading to so many unnecessary commits into the PR.

virajmehta:
This currently breaks the feedback tests [here](https://github.com/tensorzero/tensorzero/blob/1c13284eecb88bab0966603becb29a59f68e8334/gateway/tests/e2e/prometheus.rs#L167) since they don't have inferences associated (good, but should be fixed!)

gateway/src/endpoints/feedback.rs
```diff
@@ -308,6 +326,35 @@ async fn write_boolean(
...
+            };
+            return Ok(function_name);
+        }
+        MetricConfigLevel::Episode => {
+            // TODO: Implement this later as it might require db migrations.
+            return Ok("".to_string());
+        }
+    }
+}
+
```
virajmehta, line 357: I think this is a good thing to do (since demonstrations need the function name and every other kind of feedback doesn't care) but we should document it more thoroughly for future developers. 

Can you lay out the rationale more explicitly in the doc comment?

virajmehta (changes requested):
can you please add an inference write to the prometheus tests so that they don't break? 

Also might be worth adding the same check for a failure to the demonstration test in `e2e/feedback.rs`.

Otherwise, LGTM!

PR #598, opened 2024-12-25
Curate data using other kinds of metric (demonstration, float, none at all) for SFT in dashboard

virajmehta:
> [!IMPORTANT]
> Adds support for demonstration and float metrics in the dashboard, updating UI, backend, and configuration to handle new metric types.
> 
>   - **Behavior**:
>     - Adds support for `demonstration` and `float` metrics in `count.route.ts` and `MetricSelector.tsx`.
>     - Introduces `threshold` parameter for float metrics in `count.route.ts`.
>     - Updates `MetricSelector.tsx` to handle `none` metric selection.
>   - **UI**:
>     - Implements badge styling for metric properties in `MetricSelector.tsx`.
>     - Adds `onThresholdChange` handler in `MetricSelector.tsx`.
>   - **Backend**:
>     - Updates `getCuratedInferences` and `countCuratedInferences` in `clickhouse.ts` to handle new metric types.
>     - Adds functions for querying and counting demonstration and float metrics in `clickhouse.ts`.
>   - **Config**:
>     - Adds `demonstration` metric to `index.server.ts`.
>     - Updates `tensorzero.toml` to include new functions and metrics.
>   - **Tests**:
>     - Adds tests for new metric types in `clickhouse.test.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 19291bee8540d6932eca4466cfa7ae872dfb4751. It will automatically update as commits are pushed.</sup>


GabrielBianconi:
When metric is none, set placeholders to "N/A":

CleanShot 2024-12-26 at 17 23 35@2x

GabrielBianconi:
The dashboard sets "Max. Samples" to "Inferences" automatically. This is good but:

- Set it to "Curated Inferences" instead (if available)
- Set the input field's `max` to this value, so that we can't go above it.

CleanShot 2024-12-26 at 17 26 54@2x

dashboard/app/utils/clickhouse.ts
```diff
@@ -250,6 +250,19 @@ export async function countFeedbacksForMetric(
   metric_config: MetricConfig,
 ): Promise<number> {
   const metric_table_name = getMetricTableName(metric_config);
+
+  // Special handling for demonstration feedback which doesn't use metric_name
+  if (metric_config.type === "demonstration") {
+    const query = `SELECT COUNT() AS count FROM ${metric_table_name}`;
```
GabrielBianconi, line 256: You're not filtering for the function name. I don't think we have this in the table. We can do a join here for now (~ select from demonstrations, inner join with inferences, where function name = ...). Then open an issue to add this column and/or a materialized view. Right now this returns higher count than inferences with fixtures.

examples/haiku-hidden-preferences/config/tensorzero.toml
```diff
@@ -43,6 +43,17 @@ model = "gpt-4o-mini-2024-07-18"
...
+type = "chat"
+user_schema = "functions/write_haiku/user_schema.json"
+
+[functions.cheat_haiku.variants.baseline]
+weight = 1
+type = "chat_completion"
+model = "gpt-4o-2024-08-06"
+system_template = "functions/cheat_haiku/initial_prompt/system_template.minijinja"
+user_template = "functions/cheat_haiku/initial_prompt/user_template.minijinja"
+
```
GabrielBianconi, line 56: I don't think this is relevant?

GabrielBianconi (changes requested):
See above (incl. standalone comments)

GabrielBianconi:
Add a fixture for episode-level metrics

examples/haiku-hidden-preferences/config/tensorzero.toml
virajmehta: oops that's leftover from me making the fixture I'll remove

dashboard/app/utils/clickhouse.ts
virajmehta: I don't think we need to filter for function name here. We do this below in `countDemonstrationDataForFunction` and this information is already displayed

examples/haiku-hidden-preferences/config/tensorzero.toml
virajmehta: yeah idk how that snuck into this PR I was trying to not have it

GabrielBianconi (changes requested):
Almost there...

---

CleanShot 2024-12-27 at 23 37 50@2x

It should explain the issue here (<10 samples).

---

CleanShot 2024-12-27 at 23 39 09@2x

Max. Samples updates for every metric but when you select "None". (It should equal No. Inferences).

virajmehta:
I think we should instead do this: 

  useEffect(() => {
    if (counts.curatedInferenceCount !== null) {
      form.setValue(
        "maxSamples",
        Math.min(100000, counts.curatedInferenceCount),
      );
    } else if (counts.inferenceCount !== null) {
      form.setValue("maxSamples", Math.min(100000, counts.inferenceCount));
    }
  }, [counts.curatedInferenceCount, counts.inferenceCount, form]);

GabrielBianconi (approved):

Issue #335, opened 2024-09-18
Error paths are misleading for children of enums in `tensorzero.toml` config (upstream crate bug)


virajmehta:
Downstream of dtolnay/path-to-error#1.

---

**Minimal Example**

```toml
[functions.my_function]
type = "chat"

[functions.my_function.variants.my_variant]
# The following entry is mandatory but it's missing
# type = "chat_completion"
model = "anthropic::claude-3-haiku-20240307"

[gateway]
disable_observability = true
```

Right now this gives:

```
"missing field `type`\nin `functions.my_function`\n"
```

We want:

```
"missing field `type`\nin `functions.my_function.variants.my_variant`\n"
```

---

**Another Example**

We get...

```
 "Failed to load config: missing field `type`\nin `functions.play_chess_pgn`"
```

... for an error for the following TOML:

```toml
[functions.play_chess_pgn]
type = "json"
user_schema = "functions/play_chess_pgn/user_schema.json"
output_schema = "functions/play_chess_pgn/output_schema.json"

[functions.play_chess_pgn.variants.claude-3.5-sonnet]
weight = 1
type = "chat_completion"
model = "claude-3.5-sonnet-20240620"
system_template = "functions/play_chess_pgn/chess_prompt/system_template.minijinja"
user_template = "functions/play_chess_pgn/chess_prompt/user_template.minijinja"
```

despite the error being in the escaping of the name "claude-3.5-sonnet".

GabrielBianconi:
We'll pay a $500 bounty to whoever solves the upstream issue (dtolnay/path-to-error#1).

The passing criteria is a new `serde_path_to_error` release that solves our downstream issue (tensorzero/tensorzero#335).

Cabbache:
@GabrielBianconi Does [this](https://github.com/tensorzero/tensorzero/compare/main...Cabbache:tensorzero:main) work for you?

For @virajmehta 's config:
```
"Failed to load config: missing field `type`\nin `functions.play_chess_pgn`\n @ functions.play_chess_pgn.variants.claude-3.5-sonnet.weight"
```

GabrielBianconi:
Hi @Cabbache - I don't think that's correct, it's flagging `weight` but that's also not where the error is.

Here's a minimal example:

```toml
[functions.my_function]
type = "chat"

[functions.my_function.variants.my_variant]
# The following entry is mandatory but it's missing
# type = "chat_completion"
model = "anthropic::claude-3-haiku-20240307"

[...]

PR #613, opened 2024-12-26
feat: make credentials optional in test mode with warnings

ChetanXpro:
> [!IMPORTANT]
> Introduces `Credential` enum to handle optional credentials in test mode, updating provider implementations and tests accordingly.
> 
>   - **Behavior**:
>     - Introduced `Credential` enum in `model.rs` to handle credentials as `Static`, `Dynamic`, `FileContents`, `None`, or `Missing`.
>     - Updated constructors in `AnthropicProvider`, `OpenAIProvider`, and others to accept `Credential` instead of `CredentialLocation`.
>     - `get_api_key` methods now return `Option` to handle missing credentials.
>     - In test mode, missing credentials are logged as warnings instead of causing errors.
>   - **Provider Config**:
>     - `ProviderConfig` deserialization updated to convert `CredentialLocation` to `Credential` using `try_from` method.
>     - Default API key locations for `Anthropic` and `OpenAI` providers made public.
>   - **Tests**:
>     - Updated e2e tests in `fireworks.rs`, `gcp_vertex_anthropic.rs`, `gcp_vertex_gemini.rs`, and others to use new credential handling.
>     - Added logic to read credentials from environment variables and pass them to tests if available.
>   - **Misc**:
>     - Removed unused `std::env` imports from several provider files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 462f84da6c72469b6ea212e444fabda10b853900. It will automatically update as commits are pushed.</sup>


gateway/src/model.rs
```diff
@@ -346,10 +359,13 @@ impl<'de> Deserialize<'de> for ProviderConfig {
                 model_name,
                 api_base,
                 api_key_location,
-            } => ProviderConfig::OpenAI(
-                OpenAIProvider::new(model_name, api_base, api_key_location)
+            } => {
+                let key_location = api_key_location.unwrap_or(openai::default_api_key_location());
+                let credentials = convert_to_credential(key_location, "Openai",&model_name)?;
```
virajmehta, line 364: OpenAI

gateway/src/model.rs
```diff
@@ -586,6 +602,68 @@ impl<'de> Deserialize<'de> for CredentialLocation {
...
+    #[allow(unused_variables)]
+    fn try_from((location, provider_type, model_name): (CredentialLocation, &str, &str)) -> Result<Self, Self::Error> {
+        match location {
+            CredentialLocation::Env(key_name) => {
+                match env::var(key_name) {
+                    Ok(value) => Ok(Credential::Static(SecretString::from(value))),
+                    Err(_) => {
+                        #[cfg(any(test, feature = "e2e_tests"))]
+                        {
+                            warn!("You are missing the credentials required for `model.{}.provider.{}`, so the associated tests will likely fail.",model_name,provider_type);
```
virajmehta, line 627: this is a bit confusing as we use dotted paths like these for config paths (e.g. something like `models.gpt35.openai.gpt-35-turbo-0...`). Really what we want here is something like [serde_path_to_error](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/) but there's a [bug](https://github.com/tensorzero/tensorzero/issues/335) there. 

I'd suggest we do this as "You are missing the credentials required for a {provider_type} model, so the associated tests will likely fail", just so that we're not misleading people and this works in general. So it would be 

```rust
impl TryFrom<(CredentialLocation, &str)> for Credential {
    type Error = Error;
    #[allow(unused_variables)]
    fn try_from((location, provider_type): (CredentialLocation, &str)) -> Result<Self, Self::Error> {
...
```

Thanks!

gateway/src/model.rs
```diff
@@ -262,15 +268,22 @@ impl<'de> Deserialize<'de> for ProviderConfig {
         D: serde::Deserializer<'de>,
     {
         let helper = ProviderConfigHelper::deserialize(deserializer)?;
+        let convert_to_credential = |location: CredentialLocation, provider_name: &str, model_name: &str| {
```
virajmehta, line 271: if we move the conversion back into the model new() function then we can remove this.

gateway/src/inference/providers/anthropic.rs
```diff
@@ -34,7 +33,7 @@ lazy_static! {
 }
 const ANTHROPIC_API_VERSION: &str = "2023-06-01";
 
-fn default_api_key_location() -> CredentialLocation {
+pub fn default_api_key_location() -> CredentialLocation {
```
virajmehta, line 36: OK now that I see that you'll have to push these functions back outside I retract my previous statement about converting to a `Credential` prior to passing into the `new()` function. I think this becomes error prone if we have to rely on the developer getting the correct `default_api_key_location`. Can you make this private, change the interface back to `Option<CredentialLocation>` and use the `TryFrom` in each provider? We can also remove the conversion helper if we do this.

gateway/src/inference/providers/openai.rs
```diff
@@ -50,30 +49,22 @@ impl OpenAIProvider {
     pub fn new(
         model_name: String,
         api_base: Option<Url>,
-        api_key_location: Option<CredentialLocation>,
+        credentials: Credential,
```
virajmehta, line 52: change this back to credentialLocation my bad

gateway/src/inference/providers/anthropic.rs
```diff
@@ -47,28 +46,21 @@ pub struct AnthropicProvider {
 impl AnthropicProvider {
     pub fn new(
         model_name: String,
-        api_key_location: Option<CredentialLocation>,
+        credentials: Credential,
```
virajmehta, line 49: change back to CredentialLocation my bad

virajmehta: then convert

gateway/src/inference/providers/anthropic.rs
```diff
@@ -77,23 +69,26 @@ impl AnthropicProvider {
 pub enum AnthropicCredentials {
     Static(SecretString),
     Dynamic(String),
+    None
```
virajmehta, line 72: make this test only

gateway/src/model.rs
```diff
@@ -683,14 +761,14 @@ fn model_config_from_shorthand(
 ) -> Result<ModelConfig, Error> {
     let model_name = model_name.to_string();
     let provider_config = match provider_type {
-        "anthropic" => ProviderConfig::Anthropic(AnthropicProvider::new(model_name, None)?),
+        "anthropic" => ProviderConfig::Anthropic(AnthropicProvider::new(model_name, Credential::None)?),
```
virajmehta, line 764: let's change these back to `None` in line with making the changes above

virajmehta (changes requested):
Great work, I realized a few things were bad with my suggestiosn and caught a couple minor issues. I am happy to do another review in between the changes I suggested and the wider implementation. Glad you stopped for a check-in here!

ChetanXpro:
sounds good; now the flow is much clearer

ChetanXpro:
@virajmehta I was thinking - what if we implement TryFrom<Credential> for provider credentials to keep the conversion logic separate? Since I'm already working on the new() function changes, I could do this for all providers if you think it's useful. What are your thoughts?

Like:

```rust
impl TryFrom<Credential> for AnthropicCredentials {
   fn try_from(credential: Credential) -> Result<Self, Error> {
       match credential {
           Credential::Static(key) => Ok(AnthropicCredentials::Static(key)),
           // ... other matches
       }
   }
}

**Comment by virajmehta** [comment]
> @virajmehta I was thinking - what if we implement TryFrom for provider credentials to keep the conversion logic separate? Since I'm already working on the new() function changes, I could do this for all providers if you think it's useful. What are your thoughts?
> 
> Like:
> 
> ```rust
> impl TryFrom<Credential> for AnthropicCredentials {
>    fn try_from(credential: Credential) -> Result<Self, Error> {
>        match credential {
>            Credential::Static(key) => Ok(AnthropicCredentials::Static(key)),
>            // ... other matches
>        }
>    }
> }
> ```

I think this is sensible if the logic is: Pass in an optional<CredentialLocation> into Provider::new(), unwrap_or(default_location), use the TryFrom from that into a Credential, then use the TryFrom from that into a provider-specific credential. Does that match what you were thinking?

**Comment by virajmehta** [comment]
> @virajmehta I was thinking - what if we implement TryFrom for provider credentials to keep the conversion logic separate? Since I'm already working on the new() function changes, I could do this for all providers if you think it's useful. What are your thoughts?
> 
> Like:
> 
> ```rust
> impl TryFrom<Credential> for AnthropicCredentials {
>    fn try_from(credential: Credential) -> Result<Self, Error> {
>        match credential {
>            Credential::Static(key) => Ok(AnthropicCredentials::Static(key)),
>            // ... other matches
>        }
>    }
> }
> ```

I think this is sensible if the logic is: Pass in an optional<CredentialLocation> into Provider::new(), unwrap_or(default_location), use the TryFrom from that into a Credential, then use the TryFrom from that into a provider-specific credential. Does that match what you were thinking?

**Comment by ChetanXpro** [comment]
Yes, exactly! The flow would be

1. Optional location -> unwrap_or(default)
2. CredentialLocation -> Credential (using existing TryFrom)
3. Credential -> Provider-specific credential (using new TryFrom)

**Comment by virajmehta** [comment]
> Yes, exactly! The flow would be
> 
> 1. Optional location -> unwrap_or(default)
> 2. CredentialLocation -> Credential (using existing TryFrom)
> 3. Credential -> Provider-specific credential (using new TryFrom)

cool let's do that

**Comment by ChetanXpro** [comment]
> > Yes, exactly! The flow would be
> > 
> > 1. Optional location -> unwrap_or(default)
> > 2. CredentialLocation -> Credential (using existing TryFrom)
> > 3. Credential -> Provider-specific credential (using new TryFrom)
> 
> cool let's do that -- thank you for being thoughtful here!

I've updated all providers to handle optional credentials in test mode following earlier feedback. let me know if anything needs to be changed.

File: gateway/src/inference/providers/gcp_vertex_gemini.rs [hunk 9]
```diff
@@ -98,6 +69,40 @@ pub fn default_api_key_location() -> CredentialLocation {
...
+    
+    fn try_from((credentials, model): (Credential, &str)) -> Result<Self, Error> {
+        match credentials {
+            Credential::Static(value) => Ok(GCPVertexCredentials::Static(
+                GCPServiceAccountCredentials::from_path(value.expose_secret()).map_err(|e| {
+                    Error::new(ErrorDetails::GCPCredentials {
+                        message: format!("Failed to load GCP credentials: {}", e),
+                    })
+                })?,
+            )),
```
virajmehta, line 87: I think we should probably not have this -- let's just do FileContents and Dynamic for now

gateway/src/inference/providers/gcp_vertex_gemini.rs
```diff
@@ -232,6 +238,17 @@ impl GCPServiceAccountCredentials {
...
+    /// Given a path to a JSON key taken from a GCP service account, load the credentials needed to sign requests.
+    pub fn from_path(path: &str) -> Result<Self, Error> {
+        let credential_str = std::fs::read_to_string(path).map_err(|e| {
+            Error::new(ErrorDetails::GCPCredentials {
+                message: format!("Failed to read GCP Vertex Gemini credentials: {e}"),
+            })
+        })?;
+
+        Self::from_json_str(&credential_str)   
+    }
```
virajmehta, line 250: based on the suggestion above, we might be able to remove this?

virajmehta (commented):
@ChetanXpro This is looking pretty clean! made some minor comments but I think it's worth writing some tests. The main tests should be just the different TryFroms I believe. @GabrielBianconi do you think we need to add e2e tests for something like this? I think it could be good to add a test with a dynamic credential for each provider but that might be overkill

GabrielBianconi:
@virajmehta I don't think so. We need E2E tests for the "standard" case where credentials are provided, or dynamic inference is used. But not for the "don't panic on E2E tests missing credentials".

ChetanXpro:
@virajmehta  added unit tests for credential handling in all providers. each test covers:
- static credentials
- dynamic credentials
- missing credentials (test mode)
- invalid credential types and error cases

let me know if the test coverage looks good and I can move on to adding e2e tests for standard credential flows.

virajmehta:
> @virajmehta added unit tests for credential handling in all providers. each test covers:
> 
> * static credentials
> * dynamic credentials
> * missing credentials (test mode)
> * invalid credential types and error cases
> 
> let me know if the test coverage looks good and I can move on to adding e2e tests for standard credential flows.

Unit test coverage looks good! The current e2e tests already test the default heavily. What I would do there is pick a test type and change it to dynamic credentials just so that this code path is covered for each model provider. This shouldn't take too long. Thanks for all your help!

ChetanXpro:
Hey @virajmehta I have updated the inference_params_inference e2e tests to use dynamic credentials, let me know if this looks good or any changes are needed

virajmehta:
@ChetanXpro right now `cargo watch-e2e` is failing with the following output:

```
[Running 'cargo run-e2e']
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/gateway gateway/tests/e2e/tensorzero.toml`
{"timestamp":"2025-01-01T21:31:56.538013Z","level":"ERROR","message":"Error in acquiring GCP credentials: Invalid credential_location for GCPVertexAnthropic provider","spans":[{"name":"load"}]}
{"timestamp":"2025-01-01T21:31:56.538404Z","level":"ERROR","message":"Error in acquiring GCP credentials: Invalid credential_location for GCPVertexAnthropic provider\nin `models.claude-3-haiku-20240307-gcp-vertex.providers.gcp_vertex_anthropic`\n","spans":[{"name":"load"}]}
{"timestamp":"2025-01-01T21:31:56.538515Z","level":"ERROR","message":"Failed to load config: Error in acquiring GCP credentials: Invalid credential_location for GCPVertexAnthropic provider\nin `models.claude-3-haiku-20240307-gcp-vertex.providers.gcp_vertex_anthropic`\n"}
[Finished running. Exit status: 1]
```

The reason is that the default in GCP clients is:
```
pub fn default_api_key_location() -> CredentialLocation {
    CredentialLocation::Env("GCP_VERTEX_CREDENTIALS_PATH".to_string())
}
```

We should distinguish between `CredentialLocation::Env` like the key is in the env and `CredentialLocation::PathInEnv`, which is a path to a file that's specified by the environment variable. Do you mind adding that feature?

gateway/src/inference/providers/anthropic.rs
```diff
@@ -77,6 +62,24 @@ impl AnthropicProvider {
 pub enum AnthropicCredentials {
     Static(SecretString),
     Dynamic(String),
+    #[cfg(any(test, feature = "e2e_tests"))]
+    None,
+}
+
+impl TryFrom<Credential> for AnthropicCredentials {
```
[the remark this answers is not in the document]
virajmehta: let's leave them distinct for now

gateway/tests/e2e/providers/common.rs
```diff
@@ -741,6 +744,10 @@ pub async fn test_inference_params_inference_request_with_provider(provider: E2E
         "stream": false,
     });
```
virajmehta, line 745: @ChetanXpro we will need to also extend the `tensorzero.toml` file to contain a model and provider for each provider type with `credential_location = dynamic::openai_api_key` or similar. Then we'll need to make a variant that uses this model for each of the provider types and use that for inference tests.  This can't work as-is I believe (the gateway won't know where to look).

gateway/tests/e2e/providers/common.rs
```diff
@@ -21,6 +23,7 @@ pub struct E2ETestProvider {
     pub variant_name: String,
     pub model_name: String,
     pub model_provider_name: String,
+    pub credentials: Option<HashMap<String, String>>,
```
virajmehta, line 26: make this required and empty

gateway/tests/e2e/providers/common.rs
```diff
@@ -715,7 +718,7 @@ pub async fn test_simple_streaming_inference_request_with_provider(provider: E2E
 pub async fn test_inference_params_inference_request_with_provider(provider: E2ETestProvider) {
     let episode_id = Uuid::now_v7();
 
-    let payload = json!({
+    let mut payload = json!({
```
virajmehta, line 721: no need to make this mutable if we just always pass the credentials which can be empty

gateway/src/inference/providers/gcp_vertex_anthropic.rs
```diff
@@ -2225,4 +2197,57 @@ mod tests {
             }]
         );
     }
+
+    #[test]
```
virajmehta, line 2201: this test probably should go in the Gemini file

gateway/src/model.rs
```diff
@@ -586,6 +588,63 @@ impl<'de> Deserialize<'de> for CredentialLocation {
     }
```
virajmehta, line 588: ^^ up to 561. Now that we're changing the semantics of CredentialLocation so that it is being globally converted into credentials, we should distinguish between the cases where the thing in the environment is literally the key or the path to a file. Path is correct in that it is a hardcoded. But for compatibility with existing deployments we'll also want PathFromEnv(String) that reads the path from environment, then reads the file at that path during the conversion. Can you add that? We'll need it to fix GCP integrations.

gateway/tests/e2e/providers/anthropic.rs
```diff
@@ -1,37 +1,57 @@
...
 crate::generate_batch_inference_tests!(get_providers);
 
 async fn get_providers() -> E2ETestProviders {
+    let mut map = HashMap::new();
+    if let Ok(api_key) = std::env::var("ANTHROPIC_API_KEY") {
+        map.insert("ANTHROPIC_API_KEY".to_string(), api_key);
+    }
+
+    let credentials = if map.is_empty() { None } else { Some(map) };
+
```
virajmehta, line 15: I'll not repeat this comment per-provider but let's make this map empty if there's no env var. I'd use some code like 

```rust
let credentials = match std::env::var("ANTHROPIC_API_KEY") {
    Some(key) => HashMap::from([("anthropic_api_key", key)]),
    None => HashMap::new(),
};
```

virajmehta (changes requested):
See my comments, there are a few important changes to be made to the tests. But we are close!

gateway/tests/e2e/providers/common.rs
virajmehta: Note: I also prefer the lowercase snake case for the dynamic credentials as I show above since it's going to be the key in a JSON.

ChetanXpro:
Hey @viraj, I've added the PathFromEnv support to fix the GCP credential path issue. also updated the `tensorzero.toml` config for tests

gateway/tests/e2e/tensorzero.toml
```diff
@@ -133,6 +215,14 @@ routing = ["together"]
 type = "together"
 model_name = "meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo"
 
+[models."llama3.1-8b-instruct-together-dynamic"]
+routing = ["together"]
+
+[models."llama3.1-8b-instruct-together-dynamic".providers.together]
+type = "together"
+model_name = "meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo"
+credential_location = "dynamic::together_api_key"
```
virajmehta, line 224: `api_key_location`

gateway/tests/e2e/tensorzero.toml
```diff
@@ -148,13 +238,30 @@ type = "vllm"
...
 api_base = "https://pun1owldydhycl-8000.proxy.runpod.net/v1/"
 
+[models."microsoft/Phi-3.5-mini-instruct-dynamic"]
+routing = ["vllm"]
+
+[models."microsoft/Phi-3.5-mini-instruct-dynamic".providers.vllm]
+type = "vllm"
+model_name = "microsoft/Phi-3.5-mini-instruct"
+api_base = "https://pun1owldydhycl-8000.proxy.runpod.net/v1/"
+credential_location = "dynamic::vllm_api_key"
```
virajmehta, line 248: `api_key_location`

gateway/tests/e2e/tensorzero.toml
```diff
@@ -148,13 +238,30 @@ type = "vllm"
...
 type = "mistral"
 model_name = "open-mistral-nemo-2407"
 
+[models."open-mistral-nemo-2407-dynamic"]
+routing = ["mistral"]
+
+[models."open-mistral-nemo-2407-dynamic".providers.mistral]
+type = "mistral"
+model_name = "open-mistral-nemo-2407"
+credential_location = "dynamic::mistral_api_key"
```
virajmehta, line 263: `api_key_location`

gateway/tests/e2e/tensorzero.toml
```diff
@@ -169,13 +276,30 @@ routing = ["xai"]
 type = "xai"
 model_name = "grok-2-1212"
 
+[models."grok_2_1212-dynamic"]
+routing = ["xai"]
+
+[models."grok_2_1212-dynamic".providers.xai]
+type = "xai"
+model_name = "grok-2-1212"
+credential_location = "dynamic::xai_api_key"
```
virajmehta, line 285: `api_key_location`

gateway/tests/e2e/tensorzero.toml
```diff
@@ -169,13 +276,30 @@ routing = ["xai"]
...
 type = "hyperbolic"
 model_name = "meta-llama/Meta-Llama-3-70B-Instruct"
 
+[models."meta-llama/Meta-Llama-3-70B-Instruct-dynamic"]
+routing = ["hyperbolic"]
+
+[models."meta-llama/Meta-Llama-3-70B-Instruct-dynamic".providers.hyperbolic]
+type = "hyperbolic"
+model_name = "meta-llama/Meta-Llama-3-70B-Instruct"
+credential_location = "dynamic::hyperbolic_api_key"
```
virajmehta, line 300: `api_key_location`

virajmehta (changes requested):
I think this works in principle but in practice the e2e test toml file still doesn't pass validation because the field names are incorrect. Do you mind fixing that? and make sure `cargo run-e2e` works so that we don't have to go back and forth so many times. Thanks!

ChetanXpro:
Hey i have fixed this wrong field name issue also cargo run-e2e is working fine

virajmehta:
> Hey i have fixed this wrong field name issue also cargo run-e2e is working fine

awesome, I removed the GCP credential settings from the toml file since it would be super duper annoying to dynamically generate GCP tokens at test time. The tests are passing for me as well. I'll do a final review right now. Then we can merge!

virajmehta (dismissed):

gateway/src/inference/providers/fireworks.rs
```diff
@@ -48,24 +48,12 @@ impl FireworksProvider {
...
-                    .into();
-                FireworksCredentials::Static(api_key)
-            }
-            CredentialLocation::Dynamic(key_name) => FireworksCredentials::Dynamic(key_name),
-            _ => Err(Error::new(ErrorDetails::Config {
-                message: "Invalid api_key_location for Fireworks provider".to_string(),
-            }))?,
-        };
+        let credential_location = api_key_location.unwrap_or(default_api_key_location());
+        let generic_credentials = Credential::try_from((credential_location, "Fireworks"))?;
```
GabrielBianconi, line 52: We are mixing types and names (e.g. `anthropic`, `Fireworks`). Let's use types throughout, i.e. should match the `type` in config.

gateway/src/model.rs
```diff
@@ -586,6 +595,117 @@ impl<'de> Deserialize<'de> for CredentialLocation {
...
+        (location, provider_type): (CredentialLocation, &str),
+    ) -> Result<Self, Self::Error> {
+        match location {
+            CredentialLocation::Env(key_name) => match env::var(key_name) {
+                Ok(value) => Ok(Credential::Static(SecretString::from(value))),
+                Err(_) => {
+                    #[cfg(any(test, feature = "e2e_tests"))]
+                    {
+                        warn!(
+                            "You are missing the credentials required for a {} model, so the associated tests will likely fail.",
```
GabrielBianconi, line 620:
```
You are missing the credentials required for a model provider of type `{}`, so the associated tests will likely fail.
```

gateway/src/model.rs
```diff
@@ -586,6 +595,117 @@ impl<'de> Deserialize<'de> for CredentialLocation {
...
+            },
+            CredentialLocation::PathFromEnv(env_key) => {
+                // First get the path from environment variable
+                let path = match env::var(&env_key) {
+                    Ok(path) => path,
+                    Err(_) => {
+                        #[cfg(any(test, feature = "e2e_tests"))]
+                        {
+                            warn!(
+                                "Environment variable {} for {} credentials path is missing, tests will likely fail.",
```
GabrielBianconi, line 641:
```
Environment variable {} is required for a model provider of type {} but is missing, so the associated tests will likely fail.
```

gateway/src/model.rs
```diff
@@ -586,6 +595,117 @@ impl<'de> Deserialize<'de> for CredentialLocation {
...
+                    }
+                };
+                // Then read the file contents
+                match fs::read_to_string(path) {
+                    Ok(contents) => Ok(Credential::FileContents(SecretString::from(contents))),
+                    Err(e) => {
+                        #[cfg(any(test, feature = "e2e_tests"))]
+                        {
+                            warn!(
+                                "Failed to read credentials file for a {} model: {}. Tests will likely fail.",
```
GabrielBianconi, line 664:
```
Failed to read credentials file for a model provider of type {}, so the associated tests will likely fail: {}
```

gateway/src/model.rs
```diff
@@ -586,6 +595,117 @@ impl<'de> Deserialize<'de> for CredentialLocation {
...
+                    }
+                }
+            }
+            CredentialLocation::Path(path) => match fs::read_to_string(path) {
+                Ok(contents) => Ok(Credential::FileContents(SecretString::from(contents))),
+                Err(e) => {
+                    #[cfg(any(test, feature = "e2e_tests"))]
+                    {
+                        warn!(
+                            "Failed to read credentials file for a {} model: {}. Tests will likely fail.",
```
GabrielBianconi, line 687:
```
Failed to read credentials file for a model provider of type {}, so the associated tests will likely fail: {}
```

GabrielBianconi (changes requested):
If it's simple, can the error include the full provider id (`models.A.providers.B`) instead of type? Otherwise fine as is.

I also updated `CONTRIBUTING.md`.

GabrielBianconi (approved):

PR #619, opened 2024-12-26
Fix: Handle errors in prefill_json_chunk_response for non-compliant rs

GloryKim:
This PR improves the prefill_json_chunk_response function by modifying its behavior to handle errors properly. Previously, errors were created using Error::new(...) but not returned, which caused them to be silently ignored during execution. The updated implementation includes the following improvements:

Changed the function signature to Result<ProviderInferenceResponseChunk, Error> to enable error handling.
Added proper error propagation using return Err(...) where necessary:
If the content is empty, a JSON start marker { is inserted, and the function returns successfully.
If the content contains a single text block, { is prepended to the block's content.
If the content is not a text block or contains multiple blocks, an appropriate error is returned.
Enhanced code readability and maintainability by making error handling explicit.

----

> [!IMPORTANT]
> Modify `prefill_json_chunk_response` to handle errors by changing its return type to `Result` and adding error handling logic for non-compliant response formats.
> 
>   - **Behavior**:
>     - Change `prefill_json_chunk_response` signature to `Result<ProviderInferenceResponseChunk, Error>`.
>     - Handle empty content by inserting `{` and returning success.
>     - Handle single text block by prepending `{`.
>     - Return error for non-text block or multiple blocks.
>   - **Error Handling**:
>     - Use `Err(...)` to propagate errors in `prefill_json_chunk_response`.
>     - Add error messages for non-compliant response formats.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 74df7a37c869df2030b0eefedebda7e844f3950f. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gateway/src/inference/providers/anthropic.rs
+++ b/gateway/src/inference/providers/anthropic.rs
@@ -618,32 +618,42 @@ pub(crate) fn prefill_json_response(
     Ok(content)
 }
 
+/// Updated function signature to return `Result<ProviderInferenceResponseChunk, Error>`
+/// to handle errors properly instead of silently ignoring them.
 pub(crate) fn prefill_json_chunk_response(
-    chunk: ProviderInferenceResponseChunk,
-) -> ProviderInferenceResponseChunk {
-    let mut chunk = chunk;
+    mut chunk: ProviderInferenceResponseChunk,
+) -> Result<ProviderInferenceResponseChunk, Error> {
     if chunk.content.is_empty() {
+        // If the content is empty, prepend '{' to indicate the start of a JSON object.
         chunk.content = vec![ContentBlockChunk::Text(TextChunk {
             text: "{".to_string(),
             id: chunk.inference_id.to_string(),
         })];
+        return Ok(chunk);
     } else if chunk.content.len() == 1 {
+        // If there is only one content block, prepend '{' to the text.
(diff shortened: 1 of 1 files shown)


gateway/src/inference/providers/anthropic.rs
```diff
@@ -618,32 +618,42 @@ pub(crate) fn prefill_json_response(
     Ok(content)
 }
 
+/// Updated function signature to return `Result<ProviderInferenceResponseChunk, Error>`
```
virajmehta, line 621: we don't need this as a source code comment, please remove and instead maybe add a comment along the lines of:

```
/// Prefills a response with a "{" following the recommendation in 
/// Anthropic's [docs](https://docs.anthropic.com/en/docs/test-and-evaluate/strengthen-guardrails/increase-consistency#prefill-claudes-response)
```

virajmehta (changes requested):
You will definitely need to update the unit tests farther down in the file. Please do that and update the comment here. Otherwise, looks good to me.

Thanks for your contribution!

gateway/src/inference/providers/anthropic.rs
GloryKim: Thank you for reviewing the changes. Your feedback has been very helpful, and I appreciate your time and effort in reviewing this PR.

PR #629, opened 2024-12-27
Sleep in the feedback hander before validating so that we can be sure that the inferences have been written

virajmehta:
> [!IMPORTANT]
> Introduces a delay mechanism in feedback validation to ensure inferences are written, with updates to error handling and tests.
> 
>   - **Behavior**:
>     - Introduces `throttled_get_target_identifier` in `feedback.rs` to delay validation until inferences are written, using a 5-second cooldown.
>     - Replaces `get_target_identifier` with `throttled_get_target_identifier` in `write_comment`, `write_demonstration`, `write_float`, and `write_boolean` functions.
>   - **Error Handling**:
>     - Adds `InvalidUuid` error type in `error.rs` for handling invalid UUIDs.
>   - **Utilities**:
>     - Adds `uuid_elapsed` function in `uuid_util.rs` to calculate elapsed time since UUID creation.
>   - **Tests**:
>     - Updates tests in `feedback.rs` and `prometheus.rs` to remove sleep calls, relying on throttling instead.
>     - Adds tests for `uuid_elapsed` in `uuid_util.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 93c66f1c05d702a0ae061055914543acd40ce558. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gateway/src/endpoints/feedback.rs
+++ b/gateway/src/endpoints/feedback.rs
@@ -18,6 +18,11 @@ use crate::inference::types::{parse_chat_output, ContentBlock, ContentBlockOutpu
 use crate::tool::{DynamicToolParams, StaticToolConfig, ToolCall, ToolCallConfig};
 use crate::uuid_util::uuid_elapsed;
 
+/// There is a potential issue here where if we write an inference and then immediately write feedback for it,
+/// we might not be able to find the inference in the database because it hasn't been written yet.
+///
+/// This is the amount of time we want to wait after the target was supposed to have been written
+/// before we decide that the target was actually not written because we can't find it in the database.
 const FEEDBACK_COOLDOWN_PERIOD: Duration = Duration::from_secs(5);
 
 /// The expected payload is a JSON object with the following fields:
@@ -336,7 +341,7 @@ async fn write_boolean(
 
 /// This function throttles the check that an id is valid if it was created very recently.
 /// This is to avoid a race condition where the id was created (e.g. an inference was made)
-/// but the feedback is received before the id is written to the database.o
+/// but the feedback is received before the id is written to the database.
 ///
 /// The current behavior is to immediately check and return the result if the id was created
 /// more than FEEDBACK_COOLDOWN_PERIOD ago.
--- a/gateway/src/endpoints/feedback.rs
+++ b/gateway/src/endpoints/feedback.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::time::Duration;
 
 use axum::extract::State;
 use axum::{debug_handler, Json};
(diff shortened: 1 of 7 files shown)


gateway/src/endpoints/feedback.rs
```diff
@@ -15,6 +16,9 @@ use crate::function::FunctionConfig;
 use crate::gateway_util::{AppState, AppStateData, StructuredJson};
 use crate::inference::types::{parse_chat_output, ContentBlock, ContentBlockOutput, Text};
 use crate::tool::{DynamicToolParams, StaticToolConfig, ToolCall, ToolCallConfig};
+use crate::uuid_util::uuid_elapsed;
+
+const FEEDBACK_COOLDOWN_PERIOD: Duration = Duration::from_secs(5);
```
GabrielBianconi, line 21: Comment explaining

gateway/src/endpoints/feedback.rs
```diff
@@ -328,6 +334,39 @@ async fn write_boolean(
     Ok(())
 }
 
+/// This function throttles the check that an id is valid if it was created very recently.
+/// This is to avoid a race condition where the id was created (e.g. an inference was made)
+/// but the feedback is received before the id is written to the database.o
```
GabrielBianconi, line 339: `.o`

gateway/tests/e2e/prometheus.rs
```diff
@@ -315,8 +315,6 @@ async fn test_prometheus_metrics_feedback_float() {
     let response_json = response.json::<Value>().await.unwrap();
     let inference_id = response_json.get("inference_id").unwrap().as_str().unwrap();
     let inference_id = Uuid::parse_str(inference_id).unwrap();
-    // Sleep for 1 second to allow ClickHouse to write
```
GabrielBianconi, line 318: We should keep this no? Because now we're checking `/metrics` which is a different code path.

gateway/tests/e2e/feedback.rs
```diff
@@ -702,8 +697,8 @@ async fn e2e_test_boolean_feedback() {
     let response_json = response.json::<Value>().await.unwrap();
     let inference_id = response_json.get("inference_id").unwrap().as_str().unwrap();
     let inference_id = Uuid::parse_str(inference_id).unwrap();
-    // Sleep for 1 second to allow ClickHouse to write
-    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+
+    // No sleeping, we should throttle in the gateway
 
     let payload = json!({"inference_id": inference_id, "metric_name": "task_success", "value": true, "tags": {"key": tag_value, "key2": tag_value2}});
```
GabrielBianconi, line 703: For complete sanity check, let's add a check where it sleeps >cooldown first to make sure that code path is correct as well.

GabrielBianconi (changes requested):

gateway/tests/e2e/feedback.rs
virajmehta: kk I'll change one to do that

gateway/tests/e2e/prometheus.rs
virajmehta: I do not think that's why we're sleeping here but I could be wrong. These were just added yesterday.

GabrielBianconi: just realized we sleep again after, disregard

GabrielBianconi (approved):

PR #632, opened 2024-12-28
Sft progress indicator

virajmehta:


virajmehta:
Closed since error handling required some structural changes and it's better to just use that PR

Issue #230, opened 2024-08-28
Add an optional `name` field to `tools` in the config


GabrielBianconi:
Currently, given `tools.tool_name` we use `tool_name` as the name given to the model. However, this value must be globally unique, and we might want the same name for different tools.

Let's add an optional `name` field to this table. If missing, the behavior stays as is. If provided, that overrides the function name provided to the model.

hk1997:
@GabrielBianconi Can I work on this issue?

GabrielBianconi:
Sounds good, thanks @hk1997!

hk1997:
@GabrielBianconi In your original description of the issue, you are talking about a table, what is this table?

Further, I'm trying to visualize how would the toml config look like for tools after this change? And what is the actual benefit of this feature request in tensorzero?

GabrielBianconi:
@hk1997 A TOML table in the config: https://toml.io/en/v1.0.0#table

Right now we have, say:

```toml
[tools.get_temperature]
description = "Get the current temperature for a given location."
parameters = "tools/get_temperature.json"
```

But what if we wanted two tools with different parameters/description but same name? We can't today because the `tools.get_temperature` ID has to be unique. So we want:

```toml
[tools.get_temperature_A]
name = "get_temperature"
description = "Get the current temperature for a given location."
parameters = "tools/get_temperature_A.json"

[tools.get_temperature_B]
name = "get_temperature"
description = "Get the current temperature for a given location."
parameters = "tools/get_temperature_B.json"
```

This would send the tool name as `get_temperature` to the model. The behavior should be that if the `name` is provided, use that; otherwise default to the ID.

If the user tries to send two tools with the same name to the model, error.

hk1997:
In addition to updating the name in the StaticToolConfig struct , do we also want to update the key of the following hashmap to the optional name field in config struct:

```
pub struct Config<'c> {
    ...
    pub tools: HashMap<String, StaticToolConfig>,                // tool name => tool config
    ...
}
```

virajmehta:
@hk1997 we should keep the internal representation as-is, i.e. the keys in the TOML table structure should match the keys in the HashMap representation internally as they are required to be unique and this is how the deserialization works. So we shouldn't change this.

hk1997:
> @hk1997 we should keep the internal representation as-is, i.e. the keys in the TOML table structure should match the keys in the HashMap representation internally as they are required to be unique and this is how the deserialization works. So we shouldn't change this.

[...]

PR #721, opened 2025-01-07
Add optional name field to tools in the table

hk1997:
Issue: https://github.com/tensorzero/tensorzero/issues/230

Testing: Unit tests

----

> [!IMPORTANT]
> Adds optional `name` field to tool configurations, ensuring unique names and updating tests.
> 
>   - **Behavior**:
>     - Adds optional `name` field to `UninitializedToolConfig` in `config_parser.rs`.
>     - Uses `table_id` as default name if `name` is not provided.
>     - Validates uniqueness of tool names in `Config::load_from_toml()`.
>   - **Tests**:
>     - Updates `test_config_from_toml_table_valid()` to check for optional `name` field usage.
>     - Adds `test_config_from_toml_table_duplicate_tool_names()` to verify duplicate name handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 367c9298a305d0161e69240852bdd0938db9e967. It will automatically update as commits are pushed.</sup>


gateway/src/config_parser.rs
```diff
@@ -113,6 +113,17 @@ impl<'c> Config<'c> {
...
+        let mut seen_names = HashSet::new();
+        for tool_config in tools.values() {
+            let tool_name = &tool_config.name;
+            if !seen_names.insert(tool_name) {
+                return Err(ErrorDetails::Config {
+                    message: format!("Duplicate tool name: '{}' in tool configs", tool_name),
+                }
+                .into());
+            }
+        }
```
GabrielBianconi, line 126: The _name_ can be duplicated. In fact that's why we want this change - today the _ID_ (`[tools.xxx]`) can't be duplicated, but sometimes you might want to define multiple tools with the same _name_ (for different scenarios, e.g. with different schemas).

Could you remove this check and update the tests?

GabrielBianconi (changes requested):
We need to ensure+test that:

- A specific inference request can't have two tools with the same name. This should encompass static and dynamic tools. 
- The ID + name are both stored in the database.
- The correct reference (name if available, otherwise ID) is sent to the LLM.

PR #722, opened 2025-01-07
Migrate from mypy to pyright

amerinor01:
Issue: #515 

I have migrated from `MYPY` to `Pyright`. During the transition, I addressed issues where Pyright was not correctly detecting types, but I couldn't find an alternative for comprehension lists, so I ignored it for now instead.

Additionally, for the pre-commit setup, I created a .venv inside the global repository. However, I am unsure if this is the best approach and would appreciate any suggestions on best practices for this configuration.

----

> [!IMPORTANT]
> Migrate from mypy to pyright, updating configurations and handling type issues, with changes to pre-commit setup and pyright checks.
> 
>   - **Migration**:
>     - Replace `mypy` with `pyright` in `.github/workflows/general.yml` for both TensorZero and OpenAI Python clients.
>     - Add `pyright` configuration to `.pre-commit-config.yaml`.
>   - **Type Handling**:
>     - Add `# pyright: strict` to `client.py`, `types.py`, and `util.py`.
>     - Ignore type issues in list comprehensions in `types.py`.
>   - **Pre-commit Setup**:
>     - Create `.venv` for pre-commit hooks in `.pre-commit-config.yaml`.
>     - Use `bash` to activate virtual environment and run `pyright`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for edea73b85fc3de39dc8612f3a9b36fc71e9f4160. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,13 +84,6 @@ repos:
           - typescript
         pass_filenames: false
 
-      - id: pyright
-        name: pyright
-        entry: bash -c 'source .venv/bin/activate && cd clients && pyright'
-        language: python
-        types_or: [python]
-        pass_filenames: false
-
   - repo: https://github.com/srstevenson/nb-clean
     rev: 3.2.0
     hooks:
@@ -104,3 +97,10 @@ repos:
       - id: ruff
         args: ["--extend-select", "I", "--fix"]
       - id: ruff-format
+
+  - repo: https://github.com/RobertCraigie/pyright-python
+    rev: v1.1.391
+    hooks:
+    - id: pyright
+      additional_dependencies:
+        - httpx
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,6 +84,13 @@ repos:
           - typescript
         pass_filenames: false
 
+      - id: pyright
+        name: pyright
+        entry: bash -c 'cd clients && pyright'
+        language: python
+        types_or: [python]
+        pass_filenames: false
+
   - repo: https://github.com/srstevenson/nb-clean
     rev: 3.2.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -97,10 +97,20 @@ repos:
       - id: ruff
         args: ["--extend-select", "I", "--fix"]
       - id: ruff-format
-
   - repo: https://github.com/RobertCraigie/pyright-python
     rev: v1.1.391
     hooks:
-    - id: pyright
-      additional_dependencies:
-        - httpx
+      - id: pyright
+        name: "pyright: clients/python"
+        files: ^clients/python/
+        args: ["--project", "clients/python/pyproject.toml"]
+        additional_dependencies:
+          - httpx
+          - uuid_utils
+
+      - id: pyright
+        name: "pyright: clients/openai-python"
+        files: ^clients/openai-python/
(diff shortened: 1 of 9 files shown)


GabrielBianconi (changes requested):
This is pretty close, thanks!

Could you replace `# pyright: strict` with the CLI flag (GHA + pre-commit)? 

Could you also update `merge-queue.yml` and `CONTRIBUTING.md`?

amerinor01:
Hi Gabriel, Setting the CLI flag was my first idea, but copyright doesn't have that functionality, as far as I can analyze.
On the Docs, I only found three ways to set the strict flag: 
* one, set the `# pyright: strict` 
* set it at the `[tool.pyright]` section at `pyproject.toml`
* Set the configuration at `pyrightconfig.json`

>Pyright settings can also be specified in a [tool.pyright] section of a “pyproject.toml” file. A “pyrightconfig.json” file always takes precedent over “pyproject.toml” if both are present. For a sample pyproject.toml file, see [below](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#sample-pyprojecttoml-file).

> strict [array of paths, optional]: Paths of directories or files that should use “strict” analysis if they are included. This is the same as manually adding a “# pyright: strict” comment. In strict mode, most type-checking rules are enabled. Refer to [this table](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-settings-defaults) for details about which rules are enabled in strict mode. Paths may contain wildcard characters ** (a directory or multiple levels of directories), * (a sequence of zero or more characters), or ? (a single character).

https://github.com/microsoft/pyright/blob/main/docs/configuration.md

Also, I found some discussion open at the pyright repo (https://github.com/microsoft/pyright/discussions/4491) for users asking to include the CLI option, but I didn't find out if they implemented it.

amerinor01:
I decided to insert the strict configuration into  `pyproject.toml` from the clients/python directory. I guess every client should decide the level of toughness from the type checker. If you think there is a better approach, let me know!

.pre-commit-config.yaml
```diff
@@ -84,6 +84,13 @@ repos:
           - typescript
         pass_filenames: false
 
+      - id: pyright
+        name: pyright
+        entry: bash -c 'source .venv/bin/activate && cd clients && pyright'
+        language: python
+        types_or: [python]
+        pass_filenames: false
```
GabrielBianconi, line 92: Let's replace this block with the following:

https://github.com/RobertCraigie/pyright-python?tab=readme-ov-file#pre-commit

You can use `additional_dependencies` to specify the additional dependencies needed to run.

This way, `pre-commit` will ensure that the `venv` exists under the hood.

GabrielBianconi (changes requested):
Thanks, your idea looks better!

Just one final change to pre-commit and I think we we can merge. Thanks!

.pre-commit-config.yaml
GabrielBianconi: (If you haven't, run `pre-commit install` locally to make sure it runs before your local commits too.)

GabrielBianconi (approved):
@amerinor01 Made a small change to pre-commit, now it should work fine, check my commit. I think we're good to go!

@virajmehta has to merge

virajmehta (approved):

GabrielBianconi:
Thanks @amerinor01 !

PR #729, opened 2025-01-08
Creates Inference detail view & episode detail view

virajmehta:


GabrielBianconi (changes requested):
Visual:

- Inference page, format "system" card same as messages card.
- Change color in `chat_completion` to a softer color, e.g. `blue-200` in tailwind
- `baseline` and `chat_completion` pills should have same font size
CleanShot 2025-01-09 at 16 30 55@2x
- Set font size of card headers (e.g. "Basic Information", "Input") to `text-xl`. Set sub-headers inside cards (e.g. "System", "Messages") to `text-lg`. Sub-sub-headers (e.g. User/Assistant) to `text-md`.
- Put feedback table in a card like everything else.
- Put chat output in a code block too.
CleanShot 2025-01-09 at 16 33 37@2x
- Put user/assistant messages in a code block (just the message content, not the title).
- Remove :hover for feedback table badges in episode page/always.
- Is there an inference with feedback? Can't find.

GabrielBianconi (changes requested):
CleanShot 2025-01-09 at 17 48 36@2x
- Needs margin (like before), remove shadow ^

GabrielBianconi (approved):

GabrielBianconi (approved):

PR #742, opened 2025-01-10
Adding Recipe for fine-tuning based on Direct Preference Optimization (DPO) 

ankit-varma10:
### Description 
This change adds a recipe for tuning a model using OpenAI's Direct Preference Optimization (DPO) method

### Details 

- The recipe us Jaccard similarity threshold to identify the samples for which the inference was not in line with the expectation
- The recipe the pairs the inference received from the base model with the demonstration and hands it over for fine-tuning

### Testing
- The fine-tuned model shows an improvement of approximately 10% over the base model


recipes/supervised_fine_tuning/demonstrations/openai/openai_dpo.ipynb
virajmehta, line 1: let's move this out of supervised_fine_tuning as it is a separate method, maybe recipes/dpo/openai/openai_dpo.ipynb

recipes/supervised_fine_tuning/demonstrations/openai/openai_dpo.ipynb
```diff
@@ -0,0 +1,537 @@
...
+    "    )\n",
+    "\n",
+    "    # append the non-preferred output\n",
+    "    result[\"non_preferred_output\"].append(\n",
+    "        {\n",
+    "            \"role\": \"assistant\",\n",
+    "            \"content\": non_preferred_output,\n",
+    "        }\n",
+    "    )\n",
+    "\n",
```
virajmehta, line 346: I don't think this code works as-is. We store structured information in our database for input messages and outputs here. So we'll need to use similar logic as in the other recipes to render the input messages and to grab the correct output data. Here's some sample code that should do everything except for handle the two different responses:

```
def render_message(content: List[Dict[str, Any]], role: str) -> str:
    assert role in ["user", "assistant"], f"Invalid role: {role}"

    if len(content) != 1:
        raise ValueError(f"Message must have exactly one content block: {content}")

    if content[0]["type"] != "text":
        raise ValueError(f"Content block must be of type text: {content}")

    content = content[0]["value"]

    if isinstance(content, str):
        return content
    else:
        return env.render_template(role, **content)


def sample_to_openai_messages(sample) -> List[Dict[str, str]]:
    function_input = json.loads(sample["input"])

    rendered_messages = []

    # Add the system message to the rendered messages
    # If there is data passed in or a system template there must be a system message
    system = function_input.get("system", {})
    if len(system) > 0 or system_template_path:
        if system_template_path:
            system_message = env.render_template("system", **system)
            rendered_messages.append({"role": "system", "content": system_message})
        else:
            rendered_messages.append({"role": "system", "content": system})

    # Add the input messages to the rendered messages
    for message in function_input["messages"]:
        rendered_message = render_message(message["content"], message["role"])
        rendered_messages.append({"role": message["role"], "content": rendered_message})

    # Add the demonstration to the messages NOT the output here
    output = json.loads(sample["value"])

    if function_type == "chat":
        if len(output) != 1:
            raise ValueError(f"Output {output} must have exactly one content block.")

        if output[0]["type"] != "text":
            raise ValueError(f"Output {output} must be a text block.")

        rendered_messages.append({"role": "assistant", "content": output[0]["text"]})
    elif function_type == "json":
        rendered_messages.append({"role": "assistant", "content": output["raw"]})
    else:
        raise ValueError(f"Unsupported function type: {function_type}")

    return {"messages": rendered_messages}


df["openai_messages"] = df.apply(sample_to_openai_messages, axis=1)

df.head()
```
Eventually (meaning in a couple of weeks), we'll pull the common functionality out into library code but for now I would like to keep things extremely self-contained.

virajmehta (changes requested):
Generally a good start but a couple important changes: 

1. let's remove any filtering by jaccard similarity as (1) that's not needed and (2) means we can't use this for other problems. 
2. the preparation of messages is incorrect and I'd imagine will throw errors. Please use the provided sample code to correct it. 
3. let's move this into a separate directory. You'll need to write a new requirements file (can copy the existing one) and add a brief README explaining what you built.
Thanks for the contribution!

recipes/supervised_fine_tuning/demonstrations/openai/openai_dpo.ipynb
virajmehta: Hi Ankit, thanks for the question. I think this means our existing code is underdocumented honestly. If you look at the last few lines of the `render_message` impl I sent:

```
 if isinstance(content, str):
        return content
    else:
        return env.render_template(role, **content)
```
This is because TensorZero can either store messages as raw strings or as (preferentially) structured info that needs to be templated using a template. This is so that we can use a template of our choice when generating messages. Soon, we'll abstract this complexity behind a library that prepares the data more comprehensively.
So, you'll want to update the code in `render_message` and in the `function_type=='chat'` arm of `render_output_message` to use this pattern as well. I think after these changes the code will be correct!

virajmehta:
hey @ankit-varma10 made a couple small changes but this generally looks good to me at this point. However, I tried to get some results with the demonstration NER dataset and actually saw that gpt4o performs worse after DPO (doesn't necessarily even generate conformant json). Did you observe similar? I saw you got a performance increase. Was this due to the filtering by Jaccard similarity? I made sure demonstrations were only included for examples for which the inference was incorrect, i.e. `exact_match = False`.

ankit-varma10:
@virajmehta 

Thank you making the `exact_match = False` change. 

Yes, I did see the performance improve post fine-tuning. Let me share my training file, seed, and hyper-parameters. 

- Performance with base [model](https://pastebin.com/raw/DJq2Fbge), config/tensorzero.toml used for this [run](https://pastebin.com/raw/344Lhghr)
- Performance with fine-tuned [model](https://pastebin.com/raw/QtEM8Z7N), config/tensorzero.toml used for this [run](https://pastebin.com/raw/d4dZYGta)
- [File](https://pastebin.com/raw/KFZmfifY) uploaded to OpenAI for fine-tuning. 
- Seed and other params can be found [here](https://pastebin.com/raw/KFZmfifY)

I repeated the experiment a couple of times and I'm getting similar results so it should be reproducible. Let me know if I can help you with any additional information.

virajmehta:
> @virajmehta
> 
> Thank you making the `exact_match = False` change.
> 
> Yes, I did see a performance improve post fine-tuning. Let me share my training file, seed, and hyper-parameters.
> 
> * Performance with base [model](https://pastebin.com/raw/DJq2Fbge), config/tensorzero.toml used for this [run](https://pastebin.com/raw/344Lhghr)
> * Performance with fine-tuned [model](https://pastebin.com/raw/QtEM8Z7N), config/tensorzero.toml used for this [run](https://pastebin.com/raw/d4dZYGta)
> * [File](https://pastebin.com/raw/KFZmfifY) uploaded to OpenAI for fine-tuning.
> * Seed and other params can be found [here](https://pastebin.com/raw/jX9MEakp)
> 
> I repeated the experiment a couple of times and I'm getting similar results so it should be reproducible. Let me know if I can help you with any additional information.

Hey @ankit-varma10 ok yes I definitely believe you here. I will try again tomorrow to reproduce on my end and let you know if I have any issues. But assuming this works I'd be happy to merge. Thanks for the thorough documentation of your solution and the great work!

virajmehta:
@ankit-varma10 I was able to replicate your results! I think this is good to merge.

virajmehta (approved):

ankit-varma10:
@virajmehta Few questions about the development process. 

This PR was failing the ruff check. I have now fixed those. But I'm not sure if  the `pyright`  check also applies to this PR.

Do the steps listed here[](https://github.com/tensorzero/tensorzero/blob/main/CONTRIBUTING.md#python) apply to python notebooks as well? What additional checks should I run against this PR?

virajmehta:
> @virajmehta Few questions about the development process.
> 
> This PR was failing the ruff check. I have now fixed those. But I'm not sure if the `pyright` check also applies to this PR.
> 
> Do the steps listed [here](https://github.com/tensorzero/tensorzero/blob/main/CONTRIBUTING.md#python) apply to python notebooks as well? What additional checks should I run against this PR?

We run `pyright` against `clients/python` and `clients/openai-python` but not the general codebase. We run `ruff` against the whole codebase so that is why you're running into those. 

These should be enforced by `pre-commit` which is still complaining about this PR. In particular, if you run `pre-commit run --all-files` there are some changes to the `.ipynb` (and some whitespace on unrelated files 😖 ) that it asks for. 

Once you properly have `pre-commit` installed on your end you will no longer get random linter errors in CI (probably, there could be edge cases).

ankit-varma10:
@virajmehta Thanks! I think this should be good to go.

It was complaining about some whit-spaces. I fixed those and ran `pre-commit run --all-files`.

There are a bunch of whitespaces which popped up in unrelated files after changes from main was merged, but I believe this was unrelated.

virajmehta (approved):

virajmehta:
Awesome, let's merge it! Thanks for the contribution @ankit-varma10

PR #804, opened 2025-01-18
Implement UI views for all functions and for an individual function

virajmehta:
This view allows you to get the performances by variant over time for a particular function and metric. All code is functional but needs better styling. Please don't do an in-depth code review at this point but instead just offer styling suggestions so that it can get closer to a final form before we properly code review. Thanks!


GabrielBianconi:
Visual feedback:

> `/observability/functions`

* Entries in Type column → monospace font on a grey code-like badge (like things in the inference detail page > basic info)
* Count -> Inference Count

> `/observability/function/xyz`

* Basic Information > type badge > remove shadow/border (?)
* Tool Choice, etc. → use code-like format, like in inference detail page
* Parallel tool calls → true/false
* Variant Performance Over Time: chart is tool tall, make it ~2/3 of current height; limit width of individual columns if possible; add Y axis

virajmehta:
Closing as #815 covers this and also variants

PR #895, opened 2025-01-31
Get gateway in shape for new README

virajmehta:
Couple key changes: 
* add default behavior when config cannot be located (log a warning, use default / empty config)
* add model_name argument to Python and PyO3 clients
* (because I noticed it was there) updated all versions of tensorzero in examples to latest (we can now ctrl-f on each bump)
* add an example with code to reproduce what happens in new README

----

> [!IMPORTANT]
> Add default config behavior, update client interfaces, and provide new examples for TensorZero gateway.
> 
>   - **Behavior**:
>     - Add default behavior in `config_parser.rs` when config file is missing, logging a warning and using default config.
>     - Add `model_name` argument to Python and PyO3 clients in `client.py` and `lib.rs`.
>   - **Examples**:
>     - Add `examples/readme` with `README.md`, `docker-compose.yml`, `example_async_client.py`, and `example_sync_client.py` to demonstrate usage.
>   - **Dependencies**:
>     - Update `tensorzero` version to `2025.1.9` in various `requirements.txt` files across examples and recipes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for e3f55f6435ee5d28b622de5ff933f22512d9676a. It will automatically update as commits are pushed.</sup>


examples/readme/README.md
```diff
@@ -0,0 +1,33 @@
+# TensorZero README example
```
GabrielBianconi, line 1:
```
TensorZero `README` Examples
```

examples/readme/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+This directory contains the code used for the absolutely minimal example in the README at the repository root.
```
GabrielBianconi, line 3: examples

examples/readme/README.md
```diff
@@ -0,0 +1,33 @@
...
+uv venv
+uv pip sync requirements.txt
+```
+
+or
+
+```bash
+# We recommend using Python 3.10+ and a virtual environment
+pip install -r requirements.txt
+```
```
GabrielBianconi, line 25: Only Python, uv people will know

examples/readme/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+3. Run the example:
+
+```bash
+python sync_client_example.py
+python async_client_example.py
+python openai_client_example.py
```
GabrielBianconi, line 32: `example_xxx.py` so it's more obvious in file tree

examples/readme/sync_client_example.py
```diff
@@ -0,0 +1,16 @@
+from tensorzero import TensorZeroGateway
+
```
GabrielBianconi, line 2: add newline

examples/readme/README.md
GabrielBianconi, line 32: Where is `openai_client_example.py`?

GabrielBianconi (changes requested):

examples/readme/README.md
virajmehta: it's gone and removed from README because we don't support with no config yet

GabrielBianconi (approved):

PR #973, opened 2025-02-08
UI no longer needs a config file

virajmehta:
To do this, I removed the default for TENSORZERO_UI_CONFIG_PATH if not set. 

Users who haven't set this variable will have to set it in order for the UI to continue working as expected. I am open to suggestions on how to make this transition more smooth. 

I updated all docker compose files which use the UI with this variable. We will also have to update the docs when this releases to specify this. 

----

> [!IMPORTANT]
> Removes default config path for UI, requiring `TENSORZERO_UI_CONFIG_PATH` to be set, and updates `loadConfig()` to handle missing files by returning a blank config.
> 
>   - **Behavior**:
>     - Removes default for `TENSORZERO_UI_CONFIG_PATH` in `index.server.ts`, requiring users to set it for UI functionality.
>     - `loadConfig()` now checks `ENV_CONFIG_PATH` and defaults to `DEFAULT_CONFIG_PATH` if not set, returning a blank config if no file is found.
>   - **Functions**:
>     - Updates `loadConfig()` to handle missing config files by returning a blank configuration.
>     - Modifies `getConfig()` to use `ENV_CONFIG_PATH` directly.
>   - **Misc**:
>     - Updates all Docker Compose files to include `TENSORZERO_UI_CONFIG_PATH`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 9adfdb0d8e34b3f80ddec135be000ffc414d4288. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/examples/babyai/docker-compose.yml
+++ b/examples/babyai/docker-compose.yml
@@ -33,6 +33,7 @@ services:
       - ./config:/app/config:ro
     environment:
       - TENSORZERO_CLICKHOUSE_URL=http://chuser:chpassword@clickhouse:8123/tensorzero
+      - TENSORZERO_UI_CONFIG_PATH=/app/config/tensorzero.toml
     env_file:
       - ${ENV_FILE:-.env}
     ports:
--- a/examples/babyai/docker-compose.yml
+++ b/examples/babyai/docker-compose.yml
@@ -33,7 +33,6 @@ services:
       - ./config:/app/config:ro
     environment:
       - TENSORZERO_CLICKHOUSE_URL=http://chuser:chpassword@clickhouse:8123/tensorzero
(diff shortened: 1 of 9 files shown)


virajmehta:
1. if env var not set, warn
2. if env var not set but config is at `/app/config/tensorzero.toml`, use it
3. if env var not set and config is not there, use blank config
4. if env var is set and config is missing, fail

Later, we can deprecate the default for env var not being set and just use the blank config in that circumstance

GabrielBianconi (changes requested):

GabrielBianconi (approved):

PR #1008, opened 2025-02-12
Add Environment Variable Quotation Check

Kannav02:
This PR introduces a utility to check for environment variables surrounded by quotes in both the Gateway and UI components. If any quoted variables are detected, a warning will be logged to alert users to remove the quotes, preventing potential runtime issues.

Changes:

- Added check_quoted_env_var and validate_environment_variables functions in gateway/src/env_validator.rs to validate environment variables.
- Updated the main function in gateway/src/main.rs to call the validation function during startup.
- Implemented similar checks in the UI's configuration loading process.

----

> [!IMPORTANT]
> Adds utility to check and log warnings for quoted environment variables in Gateway and UI components to prevent runtime issues.
> 
>   - **Behavior**:
>     - Adds `check_quoted_env_var` and `validate_environment_variables` in `gateway/src/env_validator.rs` to log warnings for quoted environment variables.
>     - Calls `validate_environment_variables` in `gateway/src/main.rs` during startup.
>     - Implements similar environment variable checks in `ui/app/utils/config/index.server.ts`.
>   - **Misc**:
>     - Logs warnings for environment variables with quotes to prevent runtime issues.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for d85945ac8cbc21e26b67fb295c331dc80f466e14. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gateway/src/env_validator.rs
+++ b/gateway/src/env_validator.rs
@@ -1,9 +1,8 @@
-use tracing;
-
 pub fn check_quoted_env_var(key: &str) {
     if let Ok(value) = std::env::var(key) {
-        if (value.starts_with('"') && value.ends_with('"')) || 
-           (value.starts_with('\'') && value.ends_with('\'')) {
+        if (value.starts_with('"') && value.ends_with('"'))
+            || (value.starts_with('\'') && value.ends_with('\''))
+        {
             tracing::warn!(
                 "Environment variable {} contains quotes. This may cause issues. Remove the quotes from the value in your environment configuration.",
                 key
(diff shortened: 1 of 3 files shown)


GabrielBianconi (changes requested):
Thanks for the submission and apologies for the delay here.

- The gateway can be instantiated via `main.rs` but also via bindings, so the right insertion point is likely right where the variants are loaded. A grep should help you find all of them.
- We're missing other environment variables. For the gateway, make sure you tackle every provider in `providers/` as well as the ClickHouse variable.
- Same for the UI, likely best to do it when the environment vars are consumed. A grep should help you find all of them.

Thanks!

GabrielBianconi:
Closing for now, feel free to re-open/re-submit in the future.

Issue #1137, opened 2025-02-26
Copy to clipboard button in SFT results does not work


virajmehta:
To fix, you'll need to update the code in [this file](https://github.com/tensorzero/tensorzero/blob/main/ui/app/routes/optimization/supervised-fine-tuning/SFTResult.tsx) to make sure it actually does the copy to clipboard as expected

Maddoxx88:
hey @virajmehta can you assign it to me?
also how do you want the error to be handled?

GabrielBianconi:
Hi @Maddoxx88 - please go ahead and submit a PR. The "copy to clipboard" function doesn't seem to be working, so we just need to fix that.

PR #1174, opened 2025-03-01
Fix Copy to clipboard button in SFT

naotone:
Issue #1137

Following implementations were added:
- Display an error message when clipboard copying fails.
- Automatically select the text after an error occurs.

This error was caused by Secure Contexts. The navigator.clipboard.writeText function can only be executed in environments such as HTTPS or localhost. Domains like http://example.local do not support clipboard copying.
https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

document.execCommand('copy') could be used as an alternative, it was not added since it is deprecated.
The Final Result data in the screenshot was created by me for testing.

20250301-003502@2x

----

> [!IMPORTANT]
> Fixes clipboard copy error handling in `SFTResult.tsx`, displaying an error message and selecting text on failure.
> 
>   - **Behavior**:
>     - Adds error handling in `handleCopy` function in `SFTResult.tsx` for clipboard copy failures.
>     - Displays error message and selects text if clipboard copy fails.
>     - Uses `useRef` to manage text selection in `Textarea`.
>   - **UI**:
>     - Updates `Button` to use `handleCopy` for clipboard operations.
>     - Displays error message below `Button` if copy fails.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 4fa895547e4a0a822e874eaa6aeb51cfa21886bd. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/routes/optimization/supervised-fine-tuning/SFTResult.tsx
+++ b/ui/app/routes/optimization/supervised-fine-tuning/SFTResult.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useRef } from "react";
 import { Button } from "~/components/ui/button";
 import { Textarea } from "~/components/ui/textarea";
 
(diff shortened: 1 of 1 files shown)


virajmehta:
@naotone thank you for the contribution! I'm not an expert on browser features so when I built this I tested locally and thought it was good enough. I think the best UX here is just to not show the button when it's not supported. I think this is possible with [this](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#feature_detection) API:

```
if (window.isSecureContext) {
  // Page is a secure context so service workers are now available
  navigator.serviceWorker.register("/offline-worker.js").then(() => {
    // …
  });
}
```
we could probably conditionally render the button only under those circumstances. Mind making that change? Thanks

naotone:
@virajmehta Thank you for your feedback! Auto-selecting is intended to avoid manual full selection on touch devices such as tablets and mobile phones. However, since TensorZero is primarily a desktop platform, it might be fine to hide it in this case.

In the latest commit, I’ve adjusted the logic so that the copy button is only displayed when window.isSecureContext and navigator.clipboard are both available. Additionally, I’ve introduced a forceShowCopyButton prop (default: false) in case there’s a need to override this behavior.
If you’d prefer a simpler approach where the button is always hidden when unsupported, let me know, and I can make that change. Thanks!

virajmehta:
@naotone this seems like a sensible approach to me! I ran a job myself in a secure & insecure context to test. 

Thank you for the contribution!

PR #1201, opened 2025-03-04
Run pre-commit for `uv lock` and `uv export` dynamically.

GabrielBianconi:
Automatically finds all `pyproject.toml` and runs these. This way we can add `pyproject.toml` for each example/recipe/etc.

There is a bug in `uv 0.5.X`, we should to update to `0.6.X`. We can all do `0.6.4` for consistency.


GabrielBianconi:
Also ran `pre-commit` on all files and cleaned up some random issues.

virajmehta (approved):

virajmehta (approved):

GabrielBianconi:
Superseded by #1203

PR #1261, opened 2025-03-08
start parsing, storing, and returning `finish_reason`

### LINKED ISSUE #1255: Implement `finish_reason` field
From #1254.

Our OpenAI-compatible API returns `finish_reason":""` (regular + streaming). We should return the actual reason, which we're currently not parsing out of the model inferences.

We should normalize provider-specific responses as: `stop`, `length`, `content_filter`, `tool_calls` (same as OpenAI) + `unknown` (log warning).

virajmehta:
Closes #1255 

We now define a `FinishReason` enum in the types file and require an `Option<FinishReason>` in a `ProviderInferenceResponse`. 

We thread this through the cache, batch tests, and the openai compatibility endpoints as well as the `ModelInference` table. Tests have been added to Python and OpenAI clients as well as the e2e tests and some unit tests. 

----

> [!IMPORTANT]
> Introduces `FinishReason` enum to track inference completion reasons, updating models, tests, and database schema accordingly.
> 
>   - **Behavior**:
>     - Introduces `FinishReason` enum in `types.py` to track inference completion reasons.
>     - Updates `ProviderInferenceResponse` to include `Option<FinishReason>`.
>     - Integrates `FinishReason` through cache, batch tests, and OpenAI compatibility endpoints.
>   - **Database**:
>     - Adds `finish_reason` column to `ModelInference` and `ModelInferenceCache` tables via `migration_0018.rs`.
>   - **Tests**:
>     - Updates tests in `test_openai_compatibility.py` and `test_client.py` to check `finish_reason`.
>     - Adds e2e tests to verify `finish_reason` handling.
>   - **Misc**:
>     - Updates various provider files to map provider-specific stop reasons to `FinishReason`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 41a8f18cfc8458cb5d7caf6f9761ee8fcef40fa8. It will automatically update as commits are pushed.</sup>


tensorzero-internal/src/inference/providers/google_ai_studio_gemini.rs
```diff
@@ -767,9 +767,45 @@ struct GeminiResponseContent {
 }
 
 #[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+enum GeminiFinishReason {
```
GabrielBianconi, line 771: `serde::other` to some unknown entry

tensorzero-internal/src/inference/providers/gcp_vertex_gemini.rs
```diff
@@ -961,9 +962,45 @@ struct GCPVertexGeminiResponseContent {
 }
 
 #[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+enum GCPVertexGeminiFinishReason {
```
GabrielBianconi, line 966: `serde::other` to some unknown entry so we don't blow up if they introduce a new reason

GabrielBianconi (changes requested):
Make sure none of the providers can explode if they return some stop reason we're not aware of.

GabrielBianconi (approved):

PR #1339, opened 2025-03-13
Fix #1338 (Change response's `model` format in OpenAI-compatible inference endpoint)

### LINKED ISSUE #1338: Change response's `model` format in OpenAI-compatible inference endpoint
Currently, the response's `model` is the variant name (without any prefix). This can be misleading/confusing: it's not very clear what it means (and it's not "model").

We should return a special prefixed value, like we take in for the request's `model`.

If the request's `model` is `tensorzero::model_name::abc`, the response's `model` should be `tensorzero::model_name::abc`.

If the request's `model` is `tensorzero::function_name::abc`, the response's `model` should be `tensorzero::function_name::abc::variant_name::def` (or maybe just `tensorzero::variant_name::def`).

This will be a minor breaking change.

GabrielBianconi:
Currently, the response's `model` is the variant name (without any prefix). This can be misleading/confusing: it's not very clear what it means (and it's not "model").

We should return a special prefixed value, like we take in for the request's `model`.

If the request's `model` is `tensorzero::model_name::abc`, the response's `model` should be `tensorzero::model_name::abc`.

If the request's `model` is `tensorzero::function_name::abc`, the response's `model` should be `tensorzero::function_name::abc::variant_name::def` (or maybe just `tensorzero::variant_name::def`).

This will be a minor breaking change.

----

> [!IMPORTANT]
> Change response `model` format in OpenAI-compatible inference endpoint to include prefix, aligning with request format.
> 
>   - **Behavior**:
>     - Changes response `model` format in `inference_handler` in `openai_compatible.rs` to include prefix based on `function_name` or `model_name`.
>     - Updates `OpenAICompatibleResponse` and `OpenAICompatibleResponseChunk` to use new `model` format.
>   - **Tests**:
>     - Updates tests in `openai_compatibility.test.ts` and `test_openai_compatibility.py` to check for new `model` format.
>     - Adds assertions for `model` field in `openai_compatible.rs` e2e tests.
>   - **Misc**:
>     - Fixes shell command in `.pre-commit-config.yaml` for `uv-lock` and `uv-export`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 881495ced33e01ab80cbd32c77b4e991899661f5. It will automatically update as commits are pushed.</sup>


virajmehta (approved):

virajmehta (dismissed):

GabrielBianconi:
I had to make `uv-export` and `uv-lock` run for the `pyproject.toml` folder so that the relative path for the `tensorzero` dependency works.

virajmehta (approved):

PR #1363, opened 2025-03-16
Fix OpenAI content block encoding for multi-turn tool use + parallel tool use

GabrielBianconi:
We were incorrectly splitting multiple content blocks into separate messages.

----

> [!IMPORTANT]
> Fixes content block encoding in OpenAI's multi-turn and parallel tool use, ensuring correct handling and adding extensive test coverage.
> 
>   - **Behavior**:
>     - Fixes incorrect splitting of multiple content blocks into separate messages in OpenAI's multi-turn and parallel tool use.
>     - Ensures content blocks are correctly handled in `deepseek.rs` and `openai.rs`.
>   - **Tests**:
>     - Adds tests for multi-turn parallel tool calls in `openai-compatibility.test.ts`, `test_openai_compatibility.py`, and `test_client.py`.
>     - Includes e2e tests for multi-turn parallel tool use in `dicl.rs` and `common.rs`.
>   - **Misc**:
>     - Removes redundant `.cargo/config.toml` file from `python-pyo3` directory.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for a27f267aa217ccb6b82e8b4bdd9b009806611ef6. It will automatically update as commits are pushed.</sup>


tensorzero-internal/src/inference/providers/openai.rs
```diff
@@ -1044,33 +1065,185 @@ pub(super) fn tensorzero_to_openai_system_message<'a>(
     }
 }
 
-// This function returns a `Result` because of the `image.data()` - if
-// we incorrectly pass in a deserialized `Base64Image` (which will
-// have no image data), we'll produce an error.
-pub(super) fn tensorzero_to_openai_messages(
-    message: &RequestMessage,
+// // This function returns a `Result` because of the `image.data()` - if
```
virajmehta, line 1068: remove

tensorzero-internal/src/inference/providers/openai.rs
```diff
@@ -1044,33 +1065,185 @@ pub(super) fn tensorzero_to_openai_system_message<'a>(
...
+    if !user_content_blocks.is_empty() {
+        messages.push(OpenAIRequestMessage::User(OpenAIUserRequestMessage {
+            content: user_content_blocks,
+        }));
+    }
+
+    Ok(messages)
+}
+
+pub(super) fn tensorzero_to_openai_assistant_messages(
```
virajmehta, line 1233: can we just wrap these in a public method that contains the match statement on the message role?

like in the commented-out code above, there is this:
```
match message.role {
      Role::User => tensorzero_to_openai_user_messages(message),
      Role::Assistant => tensorzero_to_openai_assistant_messages(message),
};
```
We should keep this logic in the implementation and make these functions private so as to not force every implementor of `ProviderTrait` to do this.

tensorzero-internal/src/inference/providers/openai.rs
```diff
@@ -1044,33 +1065,185 @@ pub(super) fn tensorzero_to_openai_system_message<'a>(
...
+                    },
+                });
+            }
+            ContentBlock::Thought(_) => {
+                // OpenAI doesn't support thought blocks.
+                // This can only happen if the thought block was generated by another model provider.
+                // At this point, we can either convert the thought blocks to text or drop them.
+                // We chose to drop them, because it's more consistent with the behavior that OpenAI expects.
+
+                // TODO (#1361): warn when we drop thought blocks
```
virajmehta, line 1210: can you just add the log and leave implementing a test for the issue?

tensorzero-internal/tests/e2e/providers/common.rs
```diff
@@ -9255,3 +9274,629 @@ async fn check_short_inference_response(
         should_be_cached
     );
 }
+
+#[cfg(feature = "e2e_tests")]
+pub async fn test_multi_turn_parallel_tool_use_inference_request_with_provider(
+    provider: E2ETestProvider,
+) {
+    // Together's model is too dumb to figure multi-turn tool + parallel tool calls... It keeps calling the same tool over and over.
```
virajmehta, line 9282: figure out lol

virajmehta (changes requested):

virajmehta (changes requested):
Why are we changing the API? Tests and stuff look good except I think it would be worth adding an test to `test_openai_compatibility.py` that covers this case.

virajmehta (dismissed):

virajmehta (dismissed):

PR #1384, opened 2025-03-17
Fix Try with Variant button for default function 

virajmehta:
> [!IMPORTANT]
> Fixes `Try with Variant` button rendering and request preparation for `tensorzero::default` in `BasicInfo.tsx` and `VariantResponseModal.tsx`.
> 
>   - **Behavior**:
>     - `TryWithVariantButton` in `BasicInfo.tsx` is now always rendered, regardless of `inference.function_name`.
>     - In `VariantResponseModal.tsx`, requests for `tensorzero::default` are prepared using `prepareDefaultFunctionRequest()`.
>   - **Functions**:
>     - Adds `prepareDefaultFunctionRequest()` in `VariantResponseModal.tsx` to handle request preparation for `tensorzero::default`.
>     - Handles `chat` and `json` function types differently in `prepareDefaultFunctionRequest()`.
>   - **Misc**:
>     - Removes conditional rendering of `TryWithVariantButton` in `BasicInfo.tsx`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 791c815e0c2cf4b70a4e225d53eef34588178688. It will automatically update as commits are pushed.</sup>


GabrielBianconi (approved):

ui/app/routes/observability/inferences/$inference_id/VariantResponseModal.tsx
```diff
@@ -241,3 +245,32 @@ export function VariantResponseModal({
...
+  } else if (inference.function_type === "json") {
+    // This should never happen, just in case and for type safety
+    const output_schema = inference.output_schema;
+    return {
+      model_name: selectedVariant,
+      input: inference.input,
+      dryrun: true,
+      output_schema: output_schema,
+    };
+  }
```
donavon, line 275: @virajmehta 

If I were doing a code review on this PR, these would have been some of my comments:

no need for the `else` as the first `if` does a `return`. also, the function returns `undefined` here. is that what is intended?

also, in TypeScript (and JS since ES6) if the key is the same name as the variable, you can/should use the "shorthand syntax" (see `output_schema`, tool_choice, and `parallel_tool_calls` below)

also, destructure where possible.

I'm also concerned that variables with underscores are in the TypeScript codebase. This is a leaky abstraction and should be converted to/from camelCase/snakeCase at the DTO layer with a package like [json_](https://www.npmjs.com/package/json_).

```suggestion
  if (inference.function_type === "chat") {
    const { tool_params, input } = interface; // destructure
    const tool_choice = tool_params?.tool_choice;
    const parallel_tool_calls = tool_params?.parallel_tool_calls;
    const additional_tools = tool_params?.tools_available;
    
    return {
      model_name: selectedVariant,
      input,
      dryrun: true,
      tool_choice,
      parallel_tool_calls,
      // We need to add all tools as additional for the default function
      additional_tools,
    };
  }
  
  if (inference.function_type === "json") { // no need for an else
    // This should never happen, just in case and for type safety
    const { output_schema, input } = inference; // destructure
    
    return {
      model_name: selectedVariant,
      input,
      dryrun: true,
      output_schema,
    };
  }
```

ui/app/routes/observability/inferences/$inference_id/VariantResponseModal.tsx
```diff
@@ -63,12 +63,16 @@ export function VariantResponseModal({
...
+          ? prepareDefaultFunctionRequest(inference, selectedVariant)
+          : {
+              function_name: inference.function_name,
+              input: inference.input,
+              variant_name: selectedVariant,
+              dryrun: true,
+            };
+
       variantInferenceFetcher.submit(
         { data: JSON.stringify(request) },
```
donavon, line 77: why are you `JSON.stringify` the request? this seems unnecessary and something that could be unwound

virajmehta:
@donavon thanks for the comments here! I'll answer them inline but for context: I just started React in December so the feedback is much appreciated!

ui/app/routes/observability/inferences/$inference_id/VariantResponseModal.tsx
virajmehta: I had trouble getting the form data the fetcher submits to play well with nested objects --- so I figured I'd stringify and then parse / validate on the other end. If I'm missing the correct way to do it I'd love to learn.

ui/app/routes/observability/inferences/$inference_id/VariantResponseModal.tsx
virajmehta: These comments make sense to me! We make heavy use in the other parts of the project of linters like Rust's Clippy to enforce some of these style rules around destructuring, variable names, etc. I think here we'd probably like to get ESLint or similar to do the same type of thing for idiomatic TypeScript so that the fix is general across the codebase. Do you have any thoughts about this?

PR #1449, opened 2025-03-21
Implement Evals view page

virajmehta:
Adds a page for viewing eval results at /evaluations/$eval_name.

To do this, I:
* added config / database fixtures with eval data
* added config parsing for evals and tests (including returning the derived functions and metrics) as well as unit tests for that
* added 4 clickhouse queries: `getEvalRunIds`, `getEvalResults`, `getEvalStatistics`, and `countDatapointForEval` which all have integration tests. The getEvalResults function is missing tests with ragged data.
* Added migration 0020 that is currently flagged off but contains a new materialized view TagInference and other small changes to support this UI.


tensorzero-internal/src/clickhouse/migration_manager/migrations/migration_0020.rs
```diff
@@ -0,0 +1,278 @@
...
+
+        let query = r#"
+            CREATE TABLE IF NOT EXISTS TagInference
+                (
+                    key String,
+                    value String,
+                    function_name LowCardinality(String),
+                    variant_name LowCardinality(String),
+                    episode_id UUID,
+                    inference_id UUID,
```
virajmehta, line 125: do we want this to be stored as a UInt128? Might be useful for getting things in order by inference id

ui/app/utils/clickhouse/evaluations.server.ts
```diff
@@ -0,0 +1,254 @@
...
+): Promise<EvaluationRunInfo[]> {
+  const query = `
+    SELECT DISTINCT run_tag.value as eval_run_id, run_tag.variant_name as variant_name
+    FROM TagInference AS name_tag
+    INNER JOIN TagInference AS run_tag
+      ON name_tag.inference_id = run_tag.inference_id
+    WHERE name_tag.key = 'tensorzero::eval_name'
+      AND name_tag.value = {eval_name:String}
+      AND run_tag.key = 'tensorzero::eval_run_id'
+    ORDER BY eval_run_id, variant_name
```
virajmehta, line 24: should this be toUInt128(eval_run_id) DESC?

ui/app/utils/clickhouse/evaluations.server.ts
```diff
@@ -0,0 +1,254 @@
...
+    feedback.value as metric_value
+  FROM (
+    SELECT *
+    FROM {datapoint_table_name:Identifier}
+    WHERE dataset_name = {dataset_name:String}
+      AND function_name = {function_name:String}
+      AND (
+        ${staled_window_query}
+      )
+    ORDER BY id
```
virajmehta, line 77: this needs to be sorted on toUInt128. Probably means we should drop / recreate the datapoint tables. Since we haven't used them, this is OK but we'll need to flag this.

virajmehta: second thought: we should be able to do this sort in memory so not really a huge deal. Perhaps we shouldn't have included this id in the key in the first place but it's fine.

GabrielBianconi (approved):

GabrielBianconi (approved):

Issue #1132, opened 2025-02-25
Add support for image input for OpenAI fine tuning in notebook and UI


virajmehta:
[Here](https://openai.com/index/introducing-vision-to-the-fine-tuning-api/) are the docs on the topic. 

In `openai.ts` we will throw an error for fine-tuning on content blocks of type image right now. This should be fixed and similarly for the notebooks!

adityabharadwaj198:
Addressed this with PR: https://github.com/tensorzero/tensorzero/pull/1507, I'd appreciate your thoughts on it!

PR #1507, opened 2025-03-26
Added handling for images to address https://github.com/tensorzero/tensorzero/issues/1132

adityabharadwaj198:
* Addresses: https://github.com/tensorzero/tensorzero/issues/1132
* Made changes to  openai.ts to not throw error if the content is of type image. Instead it should return OpenAIMessage in correspondence with the docs (https://openai.com/index/introducing-vision-to-the-fine-tuning-api/)
* Made changes to types.ts, specifically to the OpenAIMessage class to be able to handle image types according to the docs 

----

> [!IMPORTANT]
> Adds support for image content in OpenAI fine-tuning by handling image URLs in `openai.ts` and updating `OpenAIMessage` in `types.ts`.
> 
>   - **Behavior**:
>     - `content_block_to_openai_message()` in `openai.ts` now handles `image` type by returning an `OpenAIMessage` with `image_url`.
>     - Throws error if `image_url` is missing.
>   - **Types**:
>     - `OpenAIMessage` in `types.ts` updated to include `content` as `string` or `image_url` array.
>   - **Misc**:
>     - Removes error for unsupported image content in `openai.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 9167e65e6c03c5b40d01763e55565eb68ef672d9. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/utils/supervised_fine_tuning/openai.ts
+++ b/ui/app/utils/supervised_fine_tuning/openai.ts
@@ -279,9 +279,22 @@ export function content_block_to_openai_message(
         content: content.result,
       };
     case "image":
-      throw new Error(
-        "Image content is not supported for OpenAI fine-tuning. We have an open issue for this feature at https://github.com/tensorzero/tensorzero/issues/1132.",
-      );
+      // For fine-tuning, we need to ensure the image URL is accessible
+      const imageUrl = content.image.url;
+      if (!imageUrl) {
+        throw new Error("Image URL is required for fine-tuning");
+      }
+      return {
+        role: role as OpenAIRole,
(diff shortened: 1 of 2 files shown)


GabrielBianconi:
Thank you for the submission. Unfortunately this solution won't work. TensorZero stores image inputs in object storage (e.g. S3) and the UI doesn't currently have the credentials to fetch these images. A solution would be a lot more involved with permission management (like we do in the gateway).

PR #1510, opened 2025-03-26
update mipro notebook for metric support

anndvision:
- add support for filtering demonstrations by metric
- add support for tool call and tool result message content
- update demonstration formatting


GabrielBianconi (changes requested):
- `judge_answer/baseline/user_template.minijinja` previously would have bad formatting for long multi-line predictions. Use this instead:

```
# Prediction

{{ prediction }}

# Ground Truth

{{ ground_truth }}
```

- Clarify `MAX_DEMONSTRATIONS` → `MAX_EXAMPLES_PER_DEMONSTRATION`
- Normalize mixed usage of "truth" and "ground truth" → let's use `ground_truth` everywhere?

virajmehta (changes requested):

virajmehta:
As mentioned, let's prefix all the TensorZero function names MIPRO uses with `tensorzero::optimization::mipro::` so that we completely remove any name collisions. We should also be sure and call it SIPRO in this name if we end up changing it.

recipes/mipro/mipro.ipynb
virajmehta, line 1: the readme still says that the notebook only supports demonstrations

virajmehta: I also think the "Best Performing Pair" in point 5 should change to "Best Performing Prompts" so that it could be correctly interpreted as applying to the multistep case.

virajmehta: throught -> through

virajmehta (changes requested):
couple more minor comments and we're good to merge!

virajmehta (approved):
ok I think this is good to go, LGTM

GabrielBianconi (approved):

PR #1601, opened 2025-04-02
add links to special tags to TagsTable

virajmehta:
Also added inference ids to judge feedbacks


GabrielBianconi (changes requested):
`evaluation_name` is not clickable

GabrielBianconi (approved):

PR #1686, opened 2025-04-08
Use route function props instead of `useLoaderData`

chaance:
React Router 7 now provides loader data to route modules directly via props (and types via the generated type modules), so no need for `useLoaderData`.

----

> [!IMPORTANT]
> Replace `useLoaderData` with route function props in components to utilize React Router 7's new data handling feature.
> 
>   - **Behavior**:
>     - Replace `useLoaderData` with route function props in `App`, `DatasetBuilder`, `Home`, and `VariantDetails` components.
>     - Loader data is now accessed via `loaderData` prop in `Route.ComponentProps`.
>   - **Files Affected**:
>     - `root.tsx`: `App` component updated to use `loaderData` prop.
>     - `datasets/builder/route.tsx`: `DatasetBuilder` component updated to use `loaderData` prop.
>     - `index.tsx`: `Home` component updated to use `loaderData` prop.
>     - `observability/functions/$function_name/variants/route.tsx`: `VariantDetails` component updated to use `loaderData` prop.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for b65324afff21131517b21e7c7d82b1ef8903bbf4. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/root.tsx
+++ b/ui/app/root.tsx
@@ -5,7 +5,6 @@ import {
   Outlet,
   Scripts,
   ScrollRestoration,
-  useLoaderData,
 } from "react-router";
 
 import { ConfigProvider } from "./context/config";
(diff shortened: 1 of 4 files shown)


virajmehta (approved):
LGTM -- this was something I figured out part of the way through building all this and hadn't gone back and cleaned up. Appreciate you fixing it!

PR #1687, opened 2025-04-08
Remove unnecessary forwarding of classnames

chaance:
There are many cases where components forward classnames by default to the top-level DOM element. IMO this should generally be avoided as a default practice, particularly for higher-level components where styles and behavior are more encapsulated by design.

This also over-uses the tailwind-merge utility, which is a fairly slow function that can have a real impact on slow-rendering components as the app gets larger. Normally I wouldn't worry this kind of thing until performance becomes a problem, but I've personally dealt with this particular issue many times and it's _way_ easier to squash this one early.

Didn't spend much time here, just tackled some low-hanging fruit. Happy to discuss if there are any questions.

----

> [!IMPORTANT]
> Remove unnecessary class name forwarding and reduce `tailwind-merge` usage for better encapsulation and performance.
> 
>   - **Behavior**:
>     - Removed unnecessary `className` props from components like `ActionBar`, `BasicInfoLayout`, `ContentLayout`, `PageLayout`, `SnippetLayout`, and others.
>     - Eliminated the use of `cn` function for merging class names in components like `ActionBar`, `BasicInfoLayout`, `ContentLayout`, `PageLayout`, `SnippetLayout`, and others.
>   - **Performance**:
>     - Reduced usage of `tailwind-merge` utility to improve performance in slow-rendering components.
>   - **Code Simplification**:
>     - Removed `className` prop from interfaces and components where it was not needed, such as in `ActionBar`, `BasicInfoLayout`, `ContentLayout`, `PageLayout`, `SnippetLayout`, and others.
>     - Replaced `cn` with `clsx` in `DatasetSelector.tsx` and `ModelSelector.tsx` for conditional class application.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 3d9ecd782e1a565de7f925819c06324ae058b527. It will automatically update as commits are pushed.</sup>


virajmehta (commented):
I am fine with these changes from a code perspective. @maxbokov should take a look and make sure that this isn't removing some tailwinds class that's important for design reasons. Also that you weren't planning on using those deleted snippet classes. If you're good we can merge! Thanks, @chaance

ui/app/components/inference/NewOutput.tsx
```diff
@@ -105,7 +105,7 @@ export function OutputContent({ output, outputSchema }: OutputProps) {
 
 export default function Output({ output, outputSchema }: OutputProps) {
   return (
-    <SnippetLayout className="w-full">
```
chaance, line 108: `w-full` is already applied to this component's root element, so removing should have no impact.

maxbokov (approved):
lgtm

Issue #1532, opened 2025-03-28
UI can grow wider than viewpoint


GabrielBianconi:
Certain pages can grow past the viewpoint width. This depends on the window size and the content of the page:

![Image](https://github.com/user-attachments/assets/88728dcf-8dd5-477d-9cb2-bb69fa24691b)

https://github.com/user-attachments/assets/1f33453d-909d-4587-ac76-e768a4ccec00

GabrielBianconi:
Another example: happens when you open the model inference accordion in `/observability/inferences/0195e8a9-203e-7c32-bc0d-f8347e734c23`

PR #1725, opened 2025-04-09
Prevent content layout box from growing when children cause overflow

chaance:
In cases where an element nested inside of the main content layout is larger than the content area itself, its root element will grow to accommodate the content. This is a result of the [flexbox algorithm's min-width calculation](https://www.bigbinary.com/blog/understanding-the-automatic-minimum-size-of-flex-items), and it causes the content area to grow beyond the screen.

This PR fixes that. It also removes the `ScrollArea` component in favor of simpler (and more performant) CSS to handle overflow. The Radix `ScrollArea` was introduced at a time when scrollbars were not very customizable, and in almost all cases I recommend its use. The result here should be basically the same, except the browser will use native scrollbars which respect user preferences.

Fixes https://github.com/tensorzero/tensorzero/issues/1532

Note that child containers that are scrollable (such as the Output section on an evaluation run page) can still extend beyond the content area, but its surrounding elements should no longer be effected.

----

> [!IMPORTANT]
> Fixes content overflow issue by replacing `ScrollArea` with native CSS in `ContentLayout.tsx`.
> 
>   - **Behavior**:
>     - Prevents content layout from growing when nested elements overflow in `ContentLayout.tsx`.
>     - Replaces `ScrollArea` with native CSS for handling overflow, using `overflow-hidden` and `overflow-auto`.
>   - **Removals**:
>     - Deletes `ScrollArea` component from `scroll-area.tsx`.
>   - **Fixes**:
>     - Fixes issue #1532 related to content overflow in the layout.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 317d92791946a0d24d5afdf80fb53cfc6a17f621. It will automatically update as commits are pushed.</sup>


maxbokov:
The fix looks good to me, thanks @chaance !
Idk if it should be in the scope of this PR or a separate one, but two things I've spotted that need your input after this update are the BasicInfoLayout items that must stay in overflow instead of wrapping, and Snippet Tabs that currently go out of the box instead of being hidden in the horizontal scroll (with hidden scrollbar)

image

image

chaance:
@maxbokov Yes, I noticed several issues where other components create an overflow situation. I do think those should be handled separately on a per-component basis, since this is really more about how those components collapse on small screens vs. the layout itself. The layout container _should_ be scrollable if child content forces it, but we should prevent that from happening by properly constraining those elements.

virajmehta (approved):
LGTM!

maxbokov (dismissed):

chaance:
Had to handle conflicts, looks like I need a fresh review to merge.

virajmehta (approved):

PR #1730, opened 2025-04-10
Add reusable ID and Timestamp table items 

maxbokov:
We reuse similar table components across the app in different tables. We want to isolate them instead of setting them up in every file because 1. it's easier to design and update the styles, and 2. we want to extend their functionality, which will make per-file setup problematic.

- Added TableItems — a set of common table components. Currently has two items for IDs and Timestamps;
- TableItemId takes an ID, and optionally a link, shortens the ID by default to 8 symbols, and shows the full ID on hover in the tooltip;
- TableItemTime takes a timestamp, formats the date (as we have the same formatting for all table timestamps), and prevents wrapping.

----

> [!IMPORTANT]
> Add reusable `TableItemShortUuid` and `TableItemTime` components for consistent ID and timestamp rendering in tables.
> 
>   - **Components**:
>     - Added `TableItemShortUuid` and `TableItemTime` in `TableItems.tsx` for reusable table components.
>     - `TableItemShortUuid` shortens IDs to 8 characters, shows full ID on hover, and optionally links to a URL.
>     - `TableItemTime` formats timestamps consistently and prevents wrapping.
>   - **Usage**:
>     - Replaced ID and timestamp rendering in `FeedbackTable.tsx` with `TableItemShortUuid` and `TableItemTime`.
>     - Replaced ID and timestamp rendering in `InferencesTable.tsx` with `TableItemShortUuid` and `TableItemTime`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 3de0cf3a1bb4641c460a82b5eb92efbf50b87bcb. It will automatically update as commits are pushed.</sup>


ui/app/components/ui/TableItems.tsx
```diff
@@ -0,0 +1,67 @@
...
+    );
+  },
+);
+TableItemId.displayName = "TableItemId";
+
+interface TableItemTimeProps {
+  timestamp: string | Date;
+}
+
+const TableItemTime = React.forwardRef<HTMLSpanElement, TableItemTimeProps>(
```
chaance, line 56: FYI, `forwardRef` is no longer needed in React 19. `ref` is now treated as a normal prop. Simplifies a lot of code to just get rid of it until you actually need it.

maxbokov: sg, removed. thanks!

ui/app/components/ui/TableItems.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface TableItemIdProps {
+  id: string;
+  link?: string;
+}
+
+function TableItemId({ id, link }: TableItemIdProps) {
+  const content = (
+    <span className="font-mono whitespace-nowrap">
+      {id.length > 8 ? `…${id.slice(-8)}` : id}
```
chaance, line 19: If the ID isn't truncated, I don't think it would need a tooltip since it's redundant, no?

ui/app/components/ui/TableItems.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  return (
+    <TooltipProvider delayDuration={400}>
+      <Tooltip>
+        <TooltipTrigger asChild>
+          {link ? (
+            <Link
+              to={link}
+              className="block no-underline transition-colors duration-300 hover:text-gray-500"
```
chaance, line 30:
```suggestion
              aria-label={id}
              className="block no-underline transition-colors duration-300 hover:text-gray-500"
```
Useful to label the link since the actual contents may be truncated.

ui/app/components/ui/TableItems.tsx
GabrielBianconi: Today, all IDs are UUIDs. (We use UUIDv7, so the first few characters are often very similar/identical - that's why we show the last 8 characters.)

Maybe makes sense to specialize this as `TableItemUuid` and assume truncation will always be relevant here? No strong opinion, defer to you guys.

maxbokov:
> why are we getting big formatting changes here?

hm, I haven't touch those files, it looks like pre-commit prettifier adjustment, the code is the same. Do you want me to pull main and create another pr @virajmehta ?

ui/app/components/ui/TableItems.tsx
maxbokov: Hm, yeah, I had in mind that all IDs are UUIDs and they always be truncated. I can imagine that we may have slightly different treatments for different IDs in the future based on the type of the item. Some IDs or names may be truncated on the end and not the beginning, for some of the items we may want to add icons before it, or color codes, and potentially different hover behavior. So this is more of a clean-up iteration fix for the current items.

I'd rename it to TableItemShortUuid and keep as is. If you think we should address the short IDs case anyway here, let me know @chaance , I'm down to add it.

virajmehta:
@maxbokov I'll fix these merge conflicts later today

virajmehta (approved):

GabrielBianconi (changes requested):
The ID field is not compatible with "Cmd+F". We need to overflow the text into the table (or similar), like we discussed.

maxbokov:
> The ID field is not compatible with "Cmd+F". We need to overflow the text into the table (or similar), like we discussed.

That's fair. I changed the truncation from slice to textOverflow with max-width. Looks the same, but cmd+f works now, and the full id is in the DOM. Take a look @GabrielBianconi

virajmehta (dismissed):
this looks good!  we should apply these components across the tables in the UI. Can open an issue if you want to do in another PR

GabrielBianconi (approved):

virajmehta (approved):

PR #1780, opened 2025-04-12
Handle model provider responses with no content blocks

GabrielBianconi:
This was problematic for non-streaming JSON functions because `raw` was `String` instead of `Option<String>`.

----

> [!IMPORTANT]
> Handle model provider responses with no content blocks by updating `JsonInferenceOutput` to use `Option<String>` for `raw` and adding tests for null responses.
> 
>   - **Behavior**:
>     - `JsonInferenceOutput.raw` changed from `String` to `Option<String>` in `types.py` and `mod.rs`.
>     - Updated logic in `function.rs` to handle `None` for `raw` and `parsed` outputs.
>     - Added tests for null responses in `openai-compatibility.test.ts`, `test_openai_compatibility.py`, and `test_client.py`.
>   - **Tests**:
>     - Added `test_async_chat_function_null_response` and `test_async_json_function_null_response` in `test_openai_compatibility.py`.
>     - Added `test_sync_chat_function_null_response` and `test_sync_json_function_null_response` in `test_client.py`.
>     - Added `test_chat_function_null_response` and `test_json_function_null_response` in `inference.rs`.
>   - **Misc**:
>     - Updated `SnippetContent.tsx` and `route.tsx` to handle empty content gracefully.
>     - Added `requirements.txt` to `optimizations-server`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for d950926bc2bb93f42645f8fa82f78724326a6fcf. It will automatically update as commits are pushed.</sup>


tensorzero-internal/tests/e2e/inference.rs
```diff
@@ -3015,3 +3015,75 @@ async fn test_client_adjust_tool_call() {
...
+
+    // Check Response is OK, then fields in order
+    assert_eq!(response.status(), StatusCode::OK);
+    let response_json = response.json::<Value>().await.unwrap();
+
+    // Check that the response content is an empty array (no content blocks)
+    assert!(response_json["content"].as_array().unwrap().is_empty());
+}
+
+/// Test that an json inference with null response (i.e. no generated content blocks) works as expected.
```
virajmehta, line 3049: a

tensorzero-internal/tests/e2e/inference.rs
```diff
@@ -3015,3 +3015,75 @@ async fn test_client_adjust_tool_call() {
     let last_body = { bad_gateway.last_body.lock().await.take().unwrap() };
     assert_eq!(last_body, non_stringified_tool_call_args);
 }
+
+/// Test that an json inference with null response (i.e. no generated content blocks) works as expected.
```
virajmehta, line 3019: a

virajmehta, line 3019:
```suggestion
/// Test that a json inference with null response (i.e. no generated content blocks) works as expected.
```

tensorzero-internal/tests/e2e/inference.rs
virajmehta, line 3049:
```suggestion
/// Test that a json inference with null response (i.e. no generated content blocks) works as expected.
```

ui/app/utils/clickhouse/common.ts
```diff
@@ -232,8 +232,8 @@ export const requestMessageSchema = z.object({
 export type RequestMessage = z.infer<typeof requestMessageSchema>;
 
 export const jsonInferenceOutputSchema = z.object({
-  raw: z.string(),
-  parsed: z.any().optional(),
+  raw: z.string().nullable(),
```
virajmehta, line 235: I think what you need to do here is have it default to "" instead of making this optional everwhere in the codebase. We can keep the null encoding for primary storage but for now I don't think we're gonna handle this case in the UI so I would prefer to keep the code simple.

virajmehta (changes requested):
1 request that should simplify this change

virajmehta (approved):

PR #1824, opened 2025-04-16
Refactor feedback table values

maxbokov:
As part of the redesign and refactoring of table items, we are updating the Feedback table values:

- Comments and demonstrations are truncated and have a sheet modal view on click to access the full content;
- All items have icons for better visual alignment;
- Boolean items now have an additional "unknown" state icon;

Icons logic is:
- ideally we want to have a clear signal system that gives user an instant intuition about the feedback, if it's good or bad;
- for some items, such as demonstration, there is no good/bad interpretation, so we show those in grey simply hinting the data type;
- but if the item may have good/bad state but it's not defined, we show the "unknown" state.

----

> [!IMPORTANT]
> Refactor feedback table to include icons, truncate long content with modal views, and introduce an "unknown" state for boolean items.
> 
>   - **Behavior**:
>     - Truncate comments and demonstrations in `FeedbackTable.tsx` and add modal views for full content in `FeedbackTableModal.tsx`.
>     - Add icons for all feedback items in `FeedbackValue.tsx` and `FeedbackValueItem.tsx` for better visual alignment.
>     - Introduce an "unknown" state icon for boolean items in `FeedbackValue.tsx`.
>   - **Icons**:
>     - Add `Text`, `Float`, and `QuestionMark` icons in `Icons.tsx`.
>     - Update `getFeedbackIcon()` in `icon.tsx` to include new icons and logic for feedback types.
>   - **Layout**:
>     - Update `SnippetContent` in `SnippetLayout.tsx` to handle "Content" maxHeight and adjust expansion logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for f01e88a647a204eb5622736f0e2a06ff47d77607. It will automatically update as commits are pushed.</sup>


ui/app/components/feedback/FeedbackTableModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+    </PageLayout>
+  );
+}
+
+export function DemonstrationModal({ feedback }: FeedbackTableModalProps) {
+  if (feedback.type !== "demonstration" || typeof feedback.value !== "string") {
+    return <div>Invalid demonstration feedback</div>;
+  }
+
+  return (
```
virajmehta, line 45: image
one thing I don't love here is that the content horizontally scrolls and it's hard to read for JSON output which is the common case. Can we do something where we see if the text block parses as JSON and then if it's JSON just pretty print it? I think this could probably be used widely in our codebase.

virajmehta: if you want to do this in a separate PR then please open an issue!

virajmehta (changes requested):

GabrielBianconi (approved):

virajmehta (approved):

PR #1874, opened 2025-04-21
Fix React Hooks misuse

### LINKED ISSUE #1816: `useColorAssigner` is breaking hook rules

chaance:
This PR fixes a handful of issues related to React's rules-of-hooks. All changes are simple fixes I can make with a high degree of confidence that nothing should break as a result.

I also went ahead and installed the relevant eslint plugin and enabled rules that flagged other issues that should be fixed in follow-up PRs. I want to err on the side of caution and manually test them one at a time and re-enable the linter as I go.

Closes #1816

----

> [!IMPORTANT]
> Fixes React Hooks misuse and adds eslint plugin for hooks rules enforcement.
> 
>   - **Behavior**:
>     - Fixes order of `useConfig()` call in `MetricSelector.tsx` and `SnippetLayout.tsx`.
>     - Adds `useRef` for `selectedRunIds` in `ColorAssigner.tsx` to prevent render loops.
>     - Adds comments to ignore eslint rules temporarily in multiple files.
>   - **Linting**:
>     - Installs `eslint-plugin-react-hooks` and enables it in `eslint.config.js`.
>   - **Misc**:
>     - Refactors `SupervisedFineTuning` component to separate logic and UI.
>     - Minor refactoring in `EvaluationTable.tsx` to use `useColorAssigner()` more efficiently.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 9e439cf0abc782948436d014200974f8b387a566. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/hooks/evaluations/ColorAssigner.tsx
+++ b/ui/app/hooks/evaluations/ColorAssigner.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, createContext, useContext } from "react";
+import { useRef, useState, useEffect, createContext, useContext } from "react";
 
 // Color palette with 10 distinct options
 const COLORS = [
@@ -38,6 +38,7 @@ const ColorAssignerContext = createContext<{
 }>({
   getColor: () => DEFAULT_COLOR,
 });
+ColorAssignerContext.displayName = "ColorAssignerContext";
 
 export function ColorAssignerProvider({
   children,
@@ -47,7 +48,7 @@ export function ColorAssignerProvider({
   selectedRunIds: string[];
 }) {
   // Store the mapping of runId to color
-  const [colorMap, setColorMap] = useState<Map<string, number>>(new Map());
+  const [colorMap, setColorMap] = useState(new Map<string, number>());
 
   // Function to get a color for a run ID
   const getColor = (runId: string, withHover = true) => {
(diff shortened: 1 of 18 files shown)


ui/app/hooks/evaluations/ColorAssigner.tsx
```diff
@@ -59,40 +60,55 @@ export function ColorAssignerProvider({
...
 
-    // Then assign new colors to IDs that don't have one yet
-    selectedRunIds.forEach((runId) => {
-      if (!newColorMap.has(runId)) {
-        // Find the first available color
-        let colorIndex = 0;
-        while (usedColorIndices.has(colorIndex) && colorIndex < COLORS.length) {
-          colorIndex++;
-        }
+    setColorMap((colorMap) => {
```
chaance, line 74: Using the callback signature here so we can access previous state without needing it as a dependency

virajmehta: for my learning: why would we not want to use this as a dependency/

ui/app/routes/datasets/$dataset_name/route.tsx
```diff
@@ -60,6 +60,8 @@ export default function DatasetDetailPage({
         description: `Added ${rowsAdded} rows to the dataset. Skipped ${rowsSkipped} duplicate rows.`,
       });
     }
+    // TODO: Fix and stop ignoring lint rule
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
virajmehta, line 64: also for my learning: what does this mean?

virajmehta (approved):
LGTM but I had a couple questions for my own learning. Thanks!

ui/app/hooks/evaluations/ColorAssigner.tsx
chaance: Because calling `setColorMap` will update the state itself, and if the state is a dependency that update will trigger an infinite render loop.

State updates in React are also async and batched, so it's generally best to use the callback signature whenever you set state based on its previous value to ensure that those values are correct by the time the update runs.

ui/app/routes/datasets/$dataset_name/route.tsx
chaance: It just means we should update the dependency array to include all dependencies, but I'm not sure doing that in some cases will not have unintended consequences. It's probably safe in some cases but it's also possible we can remove some of these effects altogether, so I want to hit them each one at a time to be sure.

GabrielBianconi:
(We'll merge this shortly, having some CI issues.)

PR #1919, opened 2025-04-24
Refactor action submission for VariantResponseModal

chaance:
This PR began as a minor change to fix an issue with an effect's dependency array ignoring dependencies. It became a larger refactor that eliminates the effect altogether and delegates the act of submitting to an API route to an actual event.

So one major benefit of using React Router's loader/action APIs is that it provides complementary APIs for determining the state of a given request without managing that state yourself. One flag I always look out for is when a component is managing internal state for the sole purpose of synchronizing with some other state. This inevitably leads to extra complexity and likely bugs, but it also usually comes with a [few extra `useEffect`s that probably shouldn't be there](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state).

It's important to note that `useEffect` should generally not be treated as an event handler that fires when values in the dependency array change. This is a common misconception that unfortunately comes from a lack of better APIs in React.

In this specific case, we wanted to post to an API route as soon as a dialog is opened to fire its action that fetches the dialog's data. Instead of handling this in an effect, I moved that logic into a separate function that can be called when the event that triggers the modal is fired ([this approach is recommended in the React docs](https://react.dev/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers)).

Another change here is deriving state from the fetcher itself rather than trying to synchronize it with internal state. Which brings me to another pattern I've found useful in React Router apps, which is wrapping `useFetcher` for specific route handlers. This allows you to get a little more specific with how the data is shaped in return, and wrap up submit functions to send the right arguments. I think this generally simplifies how data hooks are used.

----

> [!IMPORTANT]
> Refactor `VariantResponseModal` to use event-driven API submission, improving state management and error handling across components.
> 
>   - **Behavior**:
>     - Refactor `VariantResponseModal` to use event-driven API submission instead of `useEffect`.
>     - Introduce `useInferenceActionFetcher` in `inference.utils.tsx` for handling API requests.
>     - Derive state from fetcher instead of internal state synchronization.
>   - **Error Handling**:
>     - Add `JSONParseError` class in `common.ts` for better error handling.
>     - Improve error handling in `inference.ts` and `inference.utils.tsx`.
>   - **Components**:
>     - Update `DatapointPage` and `InferencePage` to use new fetcher logic.
>     - Remove `setIsLoading` prop from `VariantResponseModal`.
>   - **Misc**:
>     - Remove redundant `useEffect` hooks in `VariantResponseModal.tsx`.
>     - Add error boundary handling in `route.tsx` files for better user feedback.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for a92aea42b130c7ab147553c08b6ec005a33b5a6d. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/routes/datasets/$dataset_name/datapoint/$id/route.tsx
+++ b/ui/app/routes/datasets/$dataset_name/datapoint/$id/route.tsx
@@ -292,6 +292,11 @@ export default function DatapointPage({ loaderData }: Route.ComponentProps) {
     submit({ data: JSON.stringify(request) });
   };
 
+  const handleModalClose = () => {
+    setIsModalOpen(false);
+    setSelectedVariant(null);
+  };
+
   return (
     <PageLayout>
       <PageHeader label="Datapoint" name={datapoint.id} />
@@ -356,10 +361,7 @@ export default function DatapointPage({ loaderData }: Route.ComponentProps) {
           error={variantInferenceFetcher.error?.message}
           variantResponse={variantInferenceFetcher.data?.info ?? null}
           rawResponse={variantInferenceFetcher.data?.raw ?? null}
-          onClose={() => {
-            setIsModalOpen(false);
-            setSelectedVariant(null);
-          }}
+          onClose={handleModalClose}
           item={datapoint}
           selectedVariant={selectedVariant}
           source="datapoint"
--- a/ui/app/routes/datasets/$dataset_name/datapoint/$id/route.tsx
+++ b/ui/app/routes/datasets/$dataset_name/datapoint/$id/route.tsx
@@ -1,15 +1,22 @@
-import { useFetcher } from "react-router";
-import { data, isRouteErrorResponse, redirect } from "react-router";
+import type { ActionFunctionArgs } from "react-router";
+import {
+  data,
+  isRouteErrorResponse,
+  Link,
+  redirect,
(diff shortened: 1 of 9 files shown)


ui/app/routes/datasets/$dataset_name/datapoint/$id/route.tsx
```diff
@@ -345,32 +371,70 @@ export default function DatapointPage({ loaderData }: Route.ComponentProps) {
...
+        };
+    }
   }
+  return {
+    heading: "An unknown error occurred",
+    message: "Please try again later.",
+  };
+}
+
+export function ErrorBoundary({ error }: Route.ErrorBoundaryProps) {
```
chaance, line 415: This is a slightly improved error boundary that shows a more user-friendly message, though I think we can do a little better with the design.

chaance:
@virajmehta Done 👍 Let me know if anything is still unclear, happy to chat about this and improve comments as needed.

virajmehta (approved):
awesome, thanks Chance! LGTM

Issue #1764, opened 2025-04-11
HumanFeedbackModal error handling needs fixed


virajmehta:
with `VITE_TENSORZERO_UI_FF_ENABLE_DATASETS=1` if you go to [http://localhost:5173/observability/inferences/0195f845-a261-72d2-8686-774b967d938e](http://localhost:5173/observability/inferences/0195f845-a261-72d2-8686-774b967d938e) and then click `Add feedback` and select `demonstration` and edit "thinking" -> "thnking":

<img width="1199" alt="Image" src="https://github.com/user-attachments/assets/6f8177b7-7106-453e-9824-e678af9b967d" />

and submit, you get the following: 

<img width="1746" alt="Image" src="https://github.com/user-attachments/assets/9c6a0c95-cd4a-44d5-877f-89e575b971e7" />

The gateway correctly returns 400 for a JSON demonstration that fails to validate, but we should handle the error message in the dialog instead of at the page level.

Currently this is in #1758 but since this is flagged out, we can merge to `main` then fix off `main`.

PR #1945, opened 2025-04-26
`HumanFeedbackModal` form refactor

chaance:
This PR refactors the `HumanFeedbackModal` by breaking it up into a few pieces, splitting out the form into a separate component.

The main purpose of this is to decouple the underlying `Form` component, which we will ultimately replace with a fetcher form since we don't want to redirect or reload the page immediately after submission. This move will allow us to handle errors inside the dialog and redirect only if the action does not throw.

So this is a first step at ultimately closing #1764, but I wanted to keep the PR relatively small (I will stack the remaining work for that issue unless this gets merged to main first).

----

> [!IMPORTANT]
> Refactor `HumanFeedbackModal` by splitting form logic into `HumanFeedbackForm`, updating routes, and enhancing feedback handling.
> 
>   - **Refactor**:
>     - Split `HumanFeedbackModal` into `HumanFeedbackModal` and `HumanFeedbackForm` components.
>     - `HumanFeedbackForm` handles form logic, decoupling it from the modal.
>   - **Components**:
>     - `HumanFeedbackButton` now uses `ButtonProps` and spreads props.
>     - `HumanFeedbackModal` accepts `trigger` and `children` props for flexible usage.
>   - **Routes**:
>     - Updated episode and inference routes to use `HumanFeedbackForm` within `HumanFeedbackModal`.
>     - Removed `EpisodeActions.tsx` and `InferenceActions.tsx`, integrating their logic directly into the routes.
>   - **Behavior**:
>     - Form submission no longer redirects immediately; errors are handled within the dialog.
>     - Feedback form supports multiple metric types: boolean, float, comment, and demonstration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 4d1e91a35ac3f81389941cd7164c9c352dfc1863. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


ui/app/components/feedback/HumanFeedbackForm.tsx
```diff
@@ -0,0 +1,179 @@
+import { useConfig } from "~/context/config";
```
chaance, line 1: The only changes I made to this component:

- Removed the `Form` wrapper and `_action` field so we can render it directly in the route
- Added a `formError` prop and rendered it below the form to the left of the submit button

ui/app/components/feedback/HumanFeedbackModal.tsx
```diff
@@ -3,218 +3,31 @@ import {
...
-  // Exactly one of episodeId and inferenceId should be provided
-  if ((episodeId && inferenceId) || (!episodeId && !inferenceId)) {
-    throw new Error(
-      "Exactly one of episodeId and inferenceId should be provided. This is a bug. Please file a bug report at https://github.com/tensorzero/tensorzero/discussions/new?category=bug-reports",
-    );
-  }
   return (
-    <Dialog open={isOpen} onOpenChange={onClose}>
+    <Dialog open={isOpen} onOpenChange={onOpenChange}>
+      {!!trigger && <DialogTrigger asChild>{trigger}</DialogTrigger>}
```
chaance, line 24: Generally speaking I'd recommend using the dialog trigger component instead of using `onClick` to connect dialogs to triggers (there are a few usability tweaks under the hood in Radix, plus you just don't need to manage as much state in many cases).

virajmehta: I didn't know it existed!

virajmehta (dismissed):
Looks good to me ! I agree that this is substantially cleaner

virajmehta (approved):

GabrielBianconi (approved):

PR #2157, opened 2025-05-16
Show human feedback form errors in modal

chaance:
This PR makes a few changes to facilitate inline error handling in the human feedback form dialog.

- Replace `Form` with `fetcher.Form` so that submission doesn't trigger a navigation right away, letting us read the submission's state while the dialog is open
- Reset the fetcher's error when the dialog is closed so that it doesn't show up again when the user re-opens the dialog

Closes https://github.com/tensorzero/tensorzero/issues/1764

### DIFF
--- a/ui/app/components/feedback/HumanFeedbackForm.tsx
+++ b/ui/app/components/feedback/HumanFeedbackForm.tsx
@@ -16,7 +16,8 @@ import CommentFeedbackInput from "./CommentFeedbackInput";
 
 export interface HumanFeedbackFormSharedProps {
   inferenceOutput?: ContentBlockOutput[] | JsonInferenceOutput;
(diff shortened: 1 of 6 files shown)


chaance:
recheck

chaance:

chaance:
recheck

GabrielBianconi (approved):
Thanks - this is definitely better.

As a follow-up, we should get the specific error message from the gateway and display in the UI: https://github.com/tensorzero/tensorzero/issues/2168

GabrielBianconi:
Looks like this broke a Playwright test

chaance:
Yep, looking into the test failure. I'll get it fixed asap so we can merge.

PR #2236, opened 2025-05-23
Use pnpm monorepo features

chaance:
pnpm has lots of great features for JS monorepos, so this PR is an attempt to utilize them. This will enable workspace dependencies when you start adding new JS packages (Node SDK, for example).

I think some of the Docker stuff needs to be updated still. Will have to trial-and-error that a bit.


.github/workflows/general.yml
```diff
@@ -151,10 +151,16 @@ jobs:
...
         run: rustc --version
 
+      - name: Setup Node.js
+        uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e
+        with:
+          node-version: "22.9.0"
+
       - name: Install pnpm
         uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda
-        with:
```
chaance, line 156: no need for the version here since it's specified in the root `package.json`

chaance:
@virajmehta may need some assistance on the Docker front, but I do think this is worth changing to simplify development of internal packages

Aaron1011 (approved):

PR #2399, opened 2025-06-08
add shiki

chaance:
This PR adds basic syntax highlighting for JSON strings using Shiki and implements the behavior for inference params on the inference route. It uses a new `CodeBlock` component which serves to display the code but leaves parsing/formatting/highlighting to the consumer, since this is generally better handled on the server.

Before:

Screenshot 2025-06-08 at 2 59 43 PM

After:

Screenshot 2025-06-08 at 2 56 19 PM


ui/app/tailwind.css
```diff
@@ -342,18 +342,39 @@
...
+@layer components {
+  .CodeBlock {
+    &:where(.CodeBlock--with-line-numbers) code {
+      counter-reset: line;
+    }
+
+    &:where(.CodeBlock--with-line-numbers) .line {
+      counter-increment: line;
+
+      &::before {
```
chaance, line 361: This block is what adds sticky line numbers without creating a dedicated container, since that would require additional custom rendering logic with Shiki if we wanted to use our current approach.

GabrielBianconi (commented):
Looks great, thanks! There are more places that we'll want to highlight but we can do it incrementally.

LGTM but will let @virajmehta review the code before merging.

virajmehta (changes requested):
@chaance this looks good to me generally just noticed a file name is misspelled

virajmehta (approved):

PR #2504, opened 2025-06-16
Add `@tensorzero/common` package

chaance:
This PR adds an internal package for common utilities. This will be useful when we create the Tensorzero client package (as well as any other internal JS packages that get created in the future).


virajmehta (approved):

virajmehta (changes requested):
Can you move this down a level into `shared/packages`?

chaance:
@virajmehta not that it's a hard blocker but I wanted to start with this one before moving the tensorzero client code into a standalone package. Errors in particular will be useful across in both the client package and UI. Let me know where you land and if I need to make any changes.

virajmehta:
hey @chaance we settled on putting this in `/internal/packages/common` for now. Do you mind putting it there? then we can merge.

virajmehta:
Closed until we are more sure about how we want to package up our Node client

PR #2569, opened 2025-06-20
feat: add ui version logging

ChetanXpro:

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -2,6 +2,7 @@
   "name": "tensorzero-ui",
   "private": true,
   "type": "module",
+  "version": "1.0.0",
   "scripts": {
     "build": "NODE_ENV=production react-router build",
     "dev": "react-router dev",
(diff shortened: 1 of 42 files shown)


ui/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "tensorzero-ui",
   "private": true,
   "type": "module",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
ChetanXpro: Fixed the build error, removed the imports since __APP_VERSION__ works as a global with Vite's define. 

Also removed the version.ts file since we defined these variables in a global TypeScript declaration file, which we can use later to define more global variables if needed.

ui/app/utils/version.ts
```diff
@@ -5,4 +5,4 @@
  * It serves as the single source of truth for version information.
  */
 
-export const VERSION = "2025.6.3";
+export declare const __APP_VERSION__: string;
```
kincaidoneil, line 8: Can this file be deleted now since we declare `__APP_VERSION__` in globals.d.ts?

virajmehta: yes

kincaidoneil (approved):

PR #2577, opened 2025-06-20
Display file and audio messages

kincaidoneil:
For file messages, display filename, MIME type, and download and open buttons:
Screenshot 2025-06-20 at 4 18 49 PM

Show inline audio player for audio files:
Screenshot 2025-06-20 at 4 18 57 PM

Replace evaluation table tooltip with updated messages UI:
New Tooltip

----

> [!IMPORTANT]
> Enhance message display to support file and audio content with updated UI components and schemas.
> 
>   - **Behavior**:
>     - Display filename, MIME type, and download/open buttons for file messages in `SnippetContent.tsx`.
>     - Show inline audio player for audio files in `SnippetContent.tsx`.
>     - Replace evaluation table tooltip with updated messages UI in `EvaluationTable.tsx`.
>   - **Components**:
>     - Add `AudioMessage` and `FileMessage` components in `SnippetContent.tsx`.
>     - Update `InputSnippet.tsx` to handle new message types: `structured_text`, `unstructured_text`, `missing_function_text`.
>   - **Utilities**:
>     - Add `useBase64UrlToBlobUrl` hook in `use-blob-url.ts` for converting base64 to Blob URLs.
>     - Update `resolve.server.ts` to resolve new message types and file content.
>   - **Schemas**:
>     - Add new schemas for `DisplayInputMessageContent` in `common.ts`.
>     - Update `inputSchema` and related schemas to support new message types in `common.ts`.
>   - **Tests**:
>     - Update tests in `datasets.test.ts` and `evaluations.test.ts` to cover new message types and behaviors.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 2a3f455f47c536b4643931b53b15fa80bd7717a3. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #2591, opened 2025-06-22
Fix handling of unstructured and structured text messages in UI

### LINKED ISSUE #2586: UI incorrectly assumes that any text content block that parses as a JSON object is `arguments`
In the UI, if a text content block parses as a JSON object, we assume it's "arguments". But this is not necessarily true! The user could have simply sent JSON to the model.

I added a `TextIsJSON` story to #2577 covering this case.

virajmehta:
Closes #2586.

This should be merged after #2577.

This PR introduces DisplayInput* types for input, messages, and content blocks. Those types have `type="structured_text"`, `type="unstructured_text"` in them to disambiguate the cases where the database should contain a structured piece of info (arguments) vs unstructured text (text without a schema).

There are no functionality changes but this should reduce the ambiguity in our types for further UI dev.


virajmehta:
Closed in favor of #2577 which contains these changes

PR #2600, opened 2025-06-23
Add Rust optimization code in TypeScript

virajmehta:


internal/tensorzero-node/.cargo/config.toml
virajmehta, line 1: this was autogenerated by the NAPI-RS CLI not sure whether we keep it or move it or what

internal/tensorzero-node/.github/workflows/CI.yml
virajmehta, line 1: I'd be OK deleting this (I think we should actually) once we're happy with the PR. We can always go and get it from github.

tensorzero-core/src/stored_inference.rs
```diff
@@ -465,6 +468,60 @@ pub fn render_stored_inference(
     }
 }
 
+/// Since we store the input in the database in the form of ResolvedInput but without e.g. images inside,
```
virajmehta, line 471: moved from the rust client

PR #2637, opened 2025-06-25
Update dataset selector

maxbokov:
Updating dataset dropdown selector to fix current issues.

- Fixed wrapping issues and scaling, added truncation;
- Added new icons Table and TablePlus, complimentary to existing TableCheck;
- Updated "new dataset" item logic based on previous design iterations;
- Style updates for command items (global for command component) and popover (local);
- Added border-accent color.


ui/app/routes/datasets/builder/DatasetSelector.tsx
```diff
@@ -128,56 +140,94 @@ export function DatasetSelector({
...
+                            <span
+                              className={clsx(
+                                "text-sm",
+                                inputValue.trim()
+                                  ? "text-fg-primary font-mono"
+                                  : "text-fg-muted font-normal",
+                              )}
+                            >
+                              {inputValue.trim() ||
+                                "Start typing to create a new dataset..."}
```
kincaidoneil, line 160:
```suggestion
                              {inputValue.trim()}
```

Already conditionally renders this group only if `inputValue` is truthy

ui/app/routes/datasets/builder/DatasetSelector.tsx
```diff
@@ -128,56 +140,94 @@ export function DatasetSelector({
...
+                              )}
+                            />
+                            <span
+                              className={clsx(
+                                "text-sm",
+                                inputValue.trim()
+                                  ? "text-fg-primary font-mono"
+                                  : "text-fg-muted font-normal",
+                              )}
+                            >
```
kincaidoneil, line 158:
```suggestion
                            <span className="text-sm text-fg-primary font-mono">
```

ui/app/routes/datasets/builder/DatasetSelector.tsx
```diff
@@ -53,71 +53,83 @@ export function DatasetSelector({
...
+                    value={inputValue}
+                    onValueChange={handleInputChange}
+                    className="h-9"
+                  />
+                  <CommandList>
+                    <CommandEmpty className="px-4 py-2 text-sm">
+                      No datasets found.
+                    </CommandEmpty>
+                    {inputValue.trim() && (
+                      <CommandGroup>
```
kincaidoneil, line 132: I think the "Create new dataset" option should come after the existing datasets so the first existing dataset is selected immediately as they type. (That way, users can select it with the keyboard without arrow navigating down.)

ui/app/routes/datasets/builder/DatasetSelector.tsx
```diff
@@ -128,56 +140,94 @@ export function DatasetSelector({
...
-                            <span>
-                              {inputValue.trim() || "Start typing..."}
+                            <TablePlus
+                              className={clsx(
+                                "mr-2 h-4 w-4",
+                                inputValue.trim()
+                                  ? "text-blue-600"
+                                  : "text-fg-muted",
+                              )}
+                            />
```
kincaidoneil, line 150:
```suggestion
                            <TablePlus className="mr-2 h-4 w-4 text-blue-600" />
```

ui/app/routes/datasets/builder/DatasetSelector.tsx
```diff
@@ -128,56 +140,94 @@ export function DatasetSelector({
...
                           </div>
-                          <Badge
-                            variant="outline"
-                            className="bg-blue-50 text-blue-500"
+                          <span
+                            className={clsx(
+                              "text-bg-primary rounded-md bg-blue-600 px-2 py-1 text-xs",
+                              inputValue.trim() ? "font-medium" : "invisible",
+                            )}
                           >
```
kincaidoneil, line 168:
```suggestion
                          <span className="text-bg-primary rounded-md bg-blue-600 px-2 py-1 text-xs font-medium">
```

kincaidoneil (changes requested):

kincaidoneil:
And maybe we should change the create dataset option so the left hand text says something like "Create dataset `foo`" instead of the badge on the right?

See how GitHub and Notion do this, I think it's clearer

Screenshot 2025-06-26 at 2 05 14 PM

---

Screenshot 2025-06-26 at 2 08 42 PM

ui/app/routes/datasets/builder/DatasetSelector.tsx
maxbokov: It seems logical to have it appear on the top of the list, otherwise you wouldn't see it at all as the list is longer then the dropdown frame. Is it possible to have it there while selecting the first existing dataset (and you can navigate with arrow up)?

ui/app/routes/datasets/builder/DatasetSelector.tsx
```diff
@@ -128,56 +140,94 @@ export function DatasetSelector({
...
+                          key={dataset.dataset_name}
+                          value={dataset.dataset_name}
+                          onSelect={() => {
+                            field.onChange(dataset.dataset_name);
+                            setInputValue("");
+                            setOpen(false);
+                            setIsNewDataset(false);
+                          }}
+                          className="group flex w-full items-center gap-2"
+                        >
```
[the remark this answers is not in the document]
maxbokov: Good question! A border could be a bit confusing as it will conflict with the selection styles as they will hierarchically similar so you will see one item in the list with a border and one without a border but with a bg. But the feedback seems valid to me. Maybe another way to make the currently selected item more distinct is to add a light green bg to the icon, similar to function type icons. I'll try.

maxbokov:
> And maybe we should change the create dataset option so the left hand text says something like "Create dataset `foo`" instead of the badge on the right?

Love this idea @kincaidoneil ! Thanks, will do that!

virajmehta:
@maxbokov same as the other one, once @kincaidoneil approves we can merge. Fixed Python dependency issues

ui/app/routes/datasets/builder/DatasetSelector.tsx
kincaidoneil: I changed them to all be within the same group, so the library will sort the create item depending upon how well it matches vs the others. I liked the behavior when I played with it.

virajmehta (approved):

PR #2640, opened 2025-06-25
removed credential verification in tensorzero-node

virajmehta:
This is necessary before we cut a release because we don't want to require API keys in the UI


virajmehta:
Since this was `true` by mandate before, I set it to `true` everywhere except in the `tensorzero-node` package so that we won't add a new requirement to the UI

GabrielBianconi (approved):

Issue #2139, opened 2025-05-15
Add the ability to "turn off" variants from the performance plot in the function detail page.


virajmehta:
The relevant file is VariantPerformance.tsx and you can visit `/observability/functions/extract_entities` to see an example table. We should be able to "turn on/off" different variants for which data is available. I think checkboxes in a little menu that shows up on hover would be sufficient here. 

In particualr I mean for this chart: 

<img width="1320" alt="Image" src="https://github.com/user-attachments/assets/eb570e12-e0ac-4bf1-9e4b-683c041f23d8" />

there could be many variants that we have data for (like think 50 or something) and this plot would be unreadable. Since people really care about this plot it would be great if you could configure which ones show up in the chart. For now, it is fine if the filtering happens client side.

PR #2724, opened 2025-07-07
Add variant filter

chaance:
This should close https://github.com/tensorzero/tensorzero/issues/2139.

Probably want to add a few tests and stories and review the UX. This initial implementation gets the job done, but I do think it can be a little nicer if we want to do a second pass with a proper design review/iteration.


virajmehta:
Closed as already included in the playground PR a while back

PR #2794, opened 2025-07-12
Handle unsafe configuration names in the UI

### LINKED ISSUE #1898: Handle unusual function/etc. names in the UI
Imagine we have a function `[function."my_function/abc"]`. I believe the `/` will mess with the URLs in the UI.

We may need to url-encode or generate slugs for function names, variant names, and other similar elements.

GabrielBianconi:
Fix #1898.

**Every link to something named in a configuration file MUST BE ENCODED.**

### DIFF
--- a/ui/app/routes/observability/functions/$function_name/variants/VariantTemplate.tsx
+++ b/ui/app/routes/observability/functions/$function_name/variants/VariantTemplate.tsx
@@ -4,7 +4,6 @@ import {
   SnippetContent,
   SnippetTabs,
   type SnippetTab,
-  SnippetMessage,
 } from "~/components/layout/SnippetLayout";
 import { TextMessage } from "~/components/layout/SnippetContent";
 
@@ -88,12 +87,10 @@ export default function VariantTemplate({
 
             return (
               <SnippetContent maxHeight={240}>
-                <SnippetMessage>
-                  <TextMessage
-                    content={template}
-                    emptyMessage={tab?.emptyMessage}
-                  />
-                </SnippetMessage>
+                <TextMessage
+                  content={template}
+                  emptyMessage={tab?.emptyMessage}
+                />
               </SnippetContent>
             );
           }}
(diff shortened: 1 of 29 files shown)


ui/app/routes/observability/functions/$function_name/variants/VariantTemplate.tsx
```diff
@@ -119,12 +116,10 @@ export default function VariantTemplate({
         <SnippetTabs tabs={tabs} defaultTab="system_instructions">
           {() => (
             <SnippetContent maxHeight={240}>
-              <SnippetMessage>
-                <TextMessage
-                  content={content}
-                  emptyMessage="No system instructions defined."
-                />
-              </SnippetMessage>
+              <TextMessage
```
virajmehta, line 119: this seems out of place. What's this change for?

virajmehta (commented):
Ideally, we could have a lint rule or some construct that would prevent us from doing this in the future. I tried to get an LLM to do it and it didn't work so opened an issue. I don't want to block this PR on this. Before merging I want to understand the change that I commented on but otherwise, LGTM

kincaidoneil (commented):
@GabrielBianconi Thanks for taking a look at this! Sorry I didn't get to this sooner.

Looking at this I was like... _huh, it's kinda weird we have to hopefully remember to encode them everywhere ...and kinda weird we have to manually construct all these URLs to begin with. Doesn't RR have generated types for this or something?_

I dug into it, and it does export this [`href`](https://reactrouter.com/api/utils/href) util, which uses your route typegen _to type every path and parameter_.

So, I was like, let's build wrapper components to enforce fully type-checked routing - and then we can do things like encode every dynamic segment there 😎

See #2797

kincaidoneil:
Screenshot 2025-07-13 at 9 35 31 PM

I also ran into an issue where the variants weren't showing up on the function page - I think might have to do with [this Clickhouse query](https://github.com/tensorzero/tensorzero/blob/db077f4c3f329558375b21d57895ab6526b074a7/ui/app/utils/clickhouse/function.ts#L322-L346)?

virajmehta:
Holding this for #2797 and a more general solution

PR #2870, opened 2025-07-19
added function parameter to Dataset selector

virajmehta:
> [!IMPORTANT]
> Add `functionName` parameter to `DatasetSelector` for filtering datasets by function and update related logic and tests.
> 
>   - **Behavior**:
>     - Add `functionName` parameter to `DatasetSelector` in `DatasetSelector.tsx` to filter datasets by function.
>     - Update `useDatasetCounts` to accept `functionName` and modify API call accordingly.
>     - Update `LaunchEvaluationModal.tsx` to pass `functionName` to `DatasetSelector`.
>     - Disable `DatasetSelector` if `selectedEvaluationName` is not set in `LaunchEvaluationModal.tsx`.
>   - **API**:
>     - Modify `getDatasetCounts` in `datasets.server.ts` to accept `function_name` and filter results.
>     - Update `counts.route.ts` to handle `function` query parameter.
>   - **Renames**:
>     - Rename `useDatasetCountFetcher` to `useDatasetInsertCountFetcher` in `count_inserts.route.ts` and `DatasetCountDisplay.tsx`.
>   - **Tests**:
>     - Update tests in `datasets.test.ts` to cover new `function_name` filtering logic.
>   - **Misc**:
>     - Update `@tanstack/react-query` to `^5.83.0` in `package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 79c9115fb55a414d6d4d7fdd011463faf6c22d5d. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


ui/app/components/dataset/DatasetFormField.tsx
```diff
@@ -1,6 +1,6 @@
 import type { Control, FieldPath, FieldValues } from "react-hook-form";
 import { FormField, FormItem, FormLabel } from "~/components/ui/form";
-import { DatasetSelector } from "~/components/dataset/DatasetSelector";
+import DatasetSelector from "~/components/dataset/DatasetSelector";
```
chaance, line 3: Just my opinion here, but I tend to recommend folks use named exports over defaults as a rule. Named exports are generally more searchable since they are guaranteed to have a reference to the name, whereas default exports can be renamed to anything when imported. A few other reasons in some of the references below.

Not a hill to die on by any means, just my perspective! And one shared by a few others! [[1](https://x.com/mjackson/status/1602837700590989312), [2](https://www.olivare.net/blog/2023/death-to-default-exports), [3](https://webdeveloper.beehiiv.com/p/avoid-default-exports-javascript-modules), [4](https://dev.to/phuocng/avoid-using-default-exports-a1c#:~:text=Default%20exports%20can%20be%20confusing,find%20the%20functions%20they%20need.)]

ui/app/routes/api/datasets/counts.route.ts
```diff
@@ -1,16 +1,53 @@
...
+/**
+ * A hook that fetches the count of rows that would be inserted into a dataset based on the provided form values.
+ * This hook automatically refetches when the form values change.
+ *
+ * @param control - The control object from react-hook-form
+ * @returns An object containing:
+ *  - count: Number of rows that would be inserted
+ *  - isLoading: Whether the count is currently being fetched
+ */
+export function useDatasetCountFetcher(functionName: string | undefined): {
```
chaance, line 33: So this is actually a case where I think RQ would be the better choice _if you want to load this data anywhere in the route tree_. Calling `fetcher.load` in a `useEffect` is problematic for a few reasons. This does not handle cache invalidation or cancellation, and will not re-fetch when the data is updated. The React Query implementation does.

If we really wanted to get rid of React Query and avoid the footguns here, we really want to re-think how we model the data source. This means loading the data from the route on which the component that consumes it is rendered (or one of its parent routes) and passing it directly from loader data to the component via props. If the component is deeply nested and may be used in a bunch of different route contexts, React Query might actually be a better solution in that case.

My feedback when RQ was first introduced was: let's try not to have different ways of doing the same thing. I still think that's true, but I do think "load data agnostically from the route tree" is a different thing than "load data dependent on the route". That would require a bit more overhead and strictness to enforce, but ultimately a better way to think about the problem and avoid the many footguns of fetching in `useEffect`.

ui/app/routes/api/datasets/counts.route.ts
```diff
@@ -1,16 +1,53 @@
...
   try {
-    const datasetCounts = await getDatasetCounts();
+    const url = new URL(request.url);
+    const functionName = url.searchParams.get("function") ?? undefined;
+    const datasetCounts = await getDatasetCounts({
+      function_name: functionName,
+    });
     const datasets = datasetCounts.map((d) => ({
-      name: d.dataset_name,
+      dataset_name: d.dataset_name,
```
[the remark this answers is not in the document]
virajmehta: I think this was a mistake, will fix. Since the type's already a DatasetCounts no reason to duplicate the info

chaance (approved):

PR #2872, opened 2025-07-20
Add filtering by time & tags to list_inferences in Rust / Python / TS

virajmehta:


virajmehta:
Closes #2371 and #2372

PR #2933, opened 2025-07-28
[Draft] OpenAI Agents SDK Integration

bglick13:
Would be great to have first-class Agents SDK support. This is a first pass at that through patching the OpenAI client the agents SDK uses.

As a proof of concept, we patch the client with a custom version that handles `episode_id` for all API request made in the Agents SDK. A similar approach could be used to support more tensorzero features like `tensorzero::tags`,  prompt templating, and more


bglick13:

GabrielBianconi:
Thanks for the PR! We'll circle back on this soon!

PR #2975, opened 2025-08-02
use native client for inference in UI

virajmehta:
This PR replaces the typescript `fetch`-based implementation of inference with a call to the Rust client and uses autogenerated types to do so.
I was able to keep all existing Zod validation (except on the way into the inference handler, which we wouldn't want since Rust validates this and we don't want to maintain a separate Zod schema for `ClientInputParams`).
One concern is that the `extra_body` and `extra_headers` TypeScript types don't work because `ts-rs` doesn't seem to support `#[serde(transparent)]` and I couldn't figure out a way to work around.

All existing tests pass so this counts as a pure internal refactor. However, I think we should build up testing around the native client over time. This might mean we run in HTTP mode against a gateway with the `e2e_tests` feature enabled and test a variety of dummy model calls. Open to suggestions here but I don't think they have to block this PR.

### DIFF
--- a/tensorzero-core/tests/e2e/providers/common.rs
+++ b/tensorzero-core/tests/e2e/providers/common.rs
@@ -10417,32 +10417,6 @@ pub async fn test_short_inference_request_with_provider(provider: E2ETestProvide
     .await;
     tokio::time::sleep(std::time::Duration::from_millis(100)).await;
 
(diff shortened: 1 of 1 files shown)


virajmehta:
Let's hold off on merging this one until @chaance takes a look!

PR #3042, opened 2025-08-09
Improve the helm chart

contrun:
I implemented a few features to make it easier to deploy tensorzero to our kubernetes cluster.

1. Added an option to set the `imagePullSecret` to pull private container image (our private tensorzero container image is a thin layer over the original container image with the functions/models etc bundled).
2. Make mounting the configmap optionally. Since our tensorzero image has already contained the `tensorzero.toml` etc, we don't need to mount configmap any more. I made the mounting optional.
3. I prefer to set all the tensorzero runtime environment from a single secret as I don't need to update both the secret itself and the helm values options. So I added an option to export all the variables in a secret.


contrun:

contrun:
Friendly ping to @timusri to see if you have time to review the changes.

PR #3063, opened 2025-08-11
Add extra_body support for embedding models

ishbir:
This PR adds support for extra_body field in the embedding model configs. Note that it does not add corresponding support to the OpenAI compatible embeddings endpoint.

Motivation: Many embedding providers use custom body fields that are necessary to properly use their APIs.

----

> [!IMPORTANT]
> Adds `extra_body` support to embedding model configurations for custom API request fields.
> 
>   - **Behavior**:
>     - Adds `extra_body` field to `EmbeddingProviderInfo` and `EmbeddingProviderRequestInfo` in `embeddings.rs`.
>     - Updates `embed()` method in `EmbeddingProvider` trait to accept `model_provider_data` with `extra_body`.
>     - Does not add support for `extra_body` to OpenAI compatible embeddings endpoint.
>   - **Tests**:
>     - Adds `test_embedding_provider_config_with_extra_body()` in `embeddings.rs` to verify `extra_body` handling.
>     - Adds `test_embedding_extra_body()` in `openai.rs` to ensure `extra_body` modifies output dimensions.
>   - **Misc**:
>     - Adds `VOYAGE_API_KEY` to `.env.example`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for eddda92a85ea541f4f7a85e4b786e8cb5cbd3826. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/.github/workflows/merge-queue.yml
+++ b/.github/workflows/merge-queue.yml
@@ -45,6 +45,7 @@ env:
   VLLM_API_KEY: ${{ secrets.VLLM_API_KEY }}
   VLLM_API_BASE: ${{ secrets.VLLM_API_BASE }}
   VLLM_MODEL_NAME: "microsoft/Phi-3.5-mini-instruct"
+  VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
   XAI_API_KEY: ${{ secrets.XAI_API_KEY }}
 
 jobs:
--- a/examples/production-deployment/.env.example
+++ b/examples/production-deployment/.env.example
@@ -34,4 +34,5 @@ TENSORZERO_CLICKHOUSE_URL="http[s]://[username]:[password]@[hostname]:[port]/[da
 # TGI_API_KEY=example
 # TOGETHER_API_KEY=example
 # VLLM_API_KEY=example
+# VOYAGE_API_KEY=example
 # XAI_API_KEY=example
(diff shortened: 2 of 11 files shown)


ishbir:

virajmehta:
Hi @ishbir, thanks for the contribution! Will take a closer look soon but wanted to address your question: 

For the e2e test, I would ideally like to have some embedding provider that has a nonstandard API parameter that we can use to make sure this works. At a glance, the voyager [embeddings api](https://docs.voyageai.com/reference/embeddings-api) seems to fit as they use `output_dimension` as opposed to OpenAI's dimensions. So we could use `extra_body` to set `output_dimension`, set the base URL to call Voyager, and assert that the dimensions were returned as expected. Does that make sense to you? You should be able to test with your own API key and then I can install one in our CI as well.

ishbir:
Thanks, just added an E2E test and verified that it works. Had to add `#[serde(default)]` to prompt_tokens as some OpenAPI APIs (like VoyageAI) may not have it set. Also had to delete some other fields to get it working. I think the Python OpenAI client does not populate fields unless they are explicitly set (and just not defaulted) - have you considered doing the same for the OpenAI integrations with TensorZero? I wouldn't expect it to be backwards incompatible.

Aaron1011 (approved):

virajmehta:
Oops, this is a legitimate failure since I forgot to set up credentials. Will add and re-enqueue!

GabrielBianconi (approved):

PR #3065, opened 2025-08-12
Correctly handle os signals for ui container

contrun:
Currently, we are not able to properly stop the tensorzero ui container with podman (or docker) stop command.

This can be demonstrated by running podman ui in one terminal.

```
podman run --rm --network host --env TENSORZERO_GATEWAY_URL=http://127.0.0.1:3999/ --env TENSORZERO_CLICKHOUSE_URL=http://default:@127.0.0.1:8123/tensorzero --env-file .env -v "$PWD/deploy/tensorzero/":/app/config/ docker.io/tensorzero/ui
```

And then run the following command to stop the container.

```
podman ps | awk '/tensorzero\/ui/ {print $1}' | xargs -n 1 --verbose --no-run-if-empty podman stop
```

The above command will show the following output.

```
WARN[0010] StopSignal SIGTERM failed to stop container priceless_thompson in 10 seconds, resorting to SIGKILL
```

This implies that the `SIGTERM` signal was not handled by the PID 1 of the container. The reason for this is that we are spawning a subprocess `react-router-serve` in the container. The main process (which has PID 1) is still the shell, which is not responding to signals. [trapping signal from "docker stop" in bash - Stack Overflow](https://stackoverflow.com/questions/20602675/trapping-signal-from-docker-stop-in-bash) has explained this well. We need to use exec so that the UI server process can receive signals from podman/docker.

----

> [!IMPORTANT]
> Use `exec` in `ui/entrypoint.sh` to ensure `react-router-serve` handles OS signals for proper container shutdown.
> 
>   - **Behavior**:
>     - Use `exec` to launch `react-router-serve` in `ui/entrypoint.sh` to handle OS signals correctly.
>     - Ensures `SIGTERM` is received by the UI server process for proper container shutdown.
>   - **Misc**:
>     - Improves container management with `podman` and `docker` stop commands.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for aba88bad53fc4a6829b37262f5d8aedc99a99ad5. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


Aaron1011 (approved):
Thanks for the PR!

PR #3265, opened 2025-08-23
feat(helm): optionally enable config map and use all values

contrun:
- Make mounting the configmap optionally. Since our tensorzero image has already contained the tensorzero.toml etc, we don't need to mount configmap any more. I made the mounting optional.
- I prefer to set all the tensorzero runtime environment from a single secret as I don't need to update both the secret itself and the helm values options. So I added an option to export all the variables in a secret.

Originally part of https://github.com/tensorzero/tensorzero/pull/3042 . I think it is sensible to make config map improvements in a single PR.

----

> [!IMPORTANT]
> Makes config map mounting optional and allows using all environment variables from a secret in Helm charts.
> 
>   - **ConfigMap**:
>     - Makes mounting the `configMap` optional in `configmap.yaml`, `gateway.yaml`, and `ui.yaml`.
>     - Controlled by `configMap.enabled` in `values.yaml`.
>   - **Environment Variables**:
>     - Adds option to use all environment variables from a secret in `gateway.yaml` and `ui.yaml`.
>     - Controlled by `additionalEnv.secretName` and `additionalEnv.keys` in `values.yaml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for edc28e10ed59ad6c5fe5e06de5fa7ee870783403. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/examples/production-deployment-k8s-helm/templates/configmap.yaml
+++ b/examples/production-deployment-k8s-helm/templates/configmap.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.configMap.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -7,4 +8,5 @@ metadata:
 data:
   {{- range $key, $value := .Values.configMap.data }}
   {{ $key }}: {{ $value | quote }}
-  {{- end }} 
+  {{- end }}
+{{- end }}
--- a/examples/production-deployment-k8s-helm/templates/gateway.yaml
+++ b/examples/production-deployment-k8s-helm/templates/gateway.yaml
@@ -13,8 +13,10 @@ spec:
       app.kubernetes.io/component: gateway
   template:
     metadata:
+      {{- if .Values.configMap.enabled }}
       annotations:
-        checksum/config: {{ toYaml .Values.configMap | sha256sum }}
+        checksum/config: {{ toYaml .Values.configMap.data | sha256sum }}
+      {{- end }}
       labels:
         {{- include "tensorzero.selectorLabels" . | nindent 8 }}
         app.kubernetes.io/component: gateway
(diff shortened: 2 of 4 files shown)


virajmehta:
## Helm Chart Review

Thanks for this PR! The changes to make ConfigMap optional and improve secret handling look good overall. Here's my review:

### ✅ What works well
1. **ConfigMap is properly made optional** with `configMap.enabled` flag
2. **Improved secret environment handling** - the `envFrom` with `secretRef` allows importing all variables from a secret when `keys` is empty, which is more flexible
3. **Fixed checksum annotation** - using `.Values.configMap.data` instead of `.Values.configMap` is more accurate

### ⚠️ Issues to fix

1. **Bug in UI template** (lines 57-62 in `templates/ui.yaml`):
   ```yaml
   {{- if .Values.configMap.enabled }}
   volumes:
     - name: config-volume
       configMap:
         name: {{ include "tensorzero.fullname" . }}
   {{- end }}
   ```
   
   The UI template only checks `configMap.enabled` for the volumes section, but in the gateway template you check both `configMap.enabled` and `persistence.enabled`. Since UI doesn't use persistence, this is correct, but inconsistent.

2. **Missing configuration path handling**: When `configMap.enabled: false`, the gateway still uses `--config-file=/app/config/tensorzero.toml` in its args. Consider either:
   - Documenting that users need to provide config via volume mount when disabled
   - Making the config file path conditional/configurable

### 📝 Suggestions
- Consider adding a comment in `values.yaml` explaining the behavior when `configMap.enabled: false`
- The empty volumes section edge case could be handled more elegantly

Overall this is a solid improvement that adds good flexibility for configuration management! Just needs the minor fixes above.

PR #3356, opened 2025-09-01
dont construct the error unless needed in embeddings handler

virajmehta:
There were errors being logged even if they weren't being hit in the embeddings handler; we fix that here. One more reason to move towards errors being handled lazily rather than eagerly in tensorzero-core.


GabrielBianconi:
Superseded by #3359

Issue #1961, opened 2025-04-28
Enable more pedantic `clippy` lints


GabrielBianconi:
@nyurik shared the following helpful command to find violations:

```bash
cargo clippy --all-targets --workspace --message-format=json --quiet -- -W clippy::pedantic    | jq -r '.message.code.code | select(. != null and startswith("clippy::"))'    | sort | uniq -c | sort -h -r
```

In table format:

```bash
echo '|count|lint|' && echo '|---|---|' && cargo clippy --all-targets --workspace --message-format=json --quiet -- -W clippy::pedantic | jq -r '.message.code.code | select(. != null and startswith("clippy::")) | "| [" + . + "](https://rust-lang.github.io/rust-clippy/master/index.html#" + .[8:] + ") |"'     | sort | uniq -c | sort -h -r | sed 's/^/|/'
```

At the time of writing:

|count|lint|
|---|---|
|5596 | [clippy::match_same_arms](https://rust-lang.github.io/rust-clippy/master/index.html#match_same_arms) |
| 657 | [clippy::doc_markdown](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown) |
| 301 | [clippy::needless_raw_string_hashes](https://rust-lang.github.io/rust-clippy/master/index.html#needless_raw_string_hashes) |
| 282 | [clippy::missing_errors_doc](https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc) |
| 246 | [clippy::default_trait_access](https://rust-lang.github.io/rust-clippy/master/index.html#default_trait_access) |
| 201 | [clippy::must_use_candidate](https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate) |
| 136 | [clippy::redundant_closure_for_method_calls](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_for_method_calls) |
| 119 | [clippy::too_many_lines](https://rust-lang.github.io/rust-clippy/master/index.html#too_many_lines) |
| 105 | [clippy::manual_string_new](https://rust-lang.github.io/rust-clippy/master/index.html#manual_string_new) |
|  57 | [clippy::match_wildcard_for_single_variants](https://rust-lang.github.io/rust-clippy/master/index.html#match_wildcard_for_single_variants) |
|  52 | [clippy::needless_pass_by_value](https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value) |
[...]

PR #3371, opened 2025-09-02
Handle `clippy::match_same_arms`

jinnovation:
Contributes to #1961.

Some of these changes are a little questionable 😅 Let me know if there are any that we feel compromise too much on readability and we can back them out to replace with `#[expect(match_same_arms)]`


GabrielBianconi:
I think this lint might be _too pedantic_ and we should leave it to the discretion of the authors (along with `single_match_else`). Readability looks so much worse to me, for questionable benefit in most cases...

Updated the PR to show we'll explicitly allow these two lints. Thanks!

PR #3399, opened 2025-09-03
Add pydantic integration example

slbotbm:


slbotbm:
I apologize, seems like I made a mistake.

PR #3470, opened 2025-09-09
Validate the DICL variants from the database instead of config

### LINKED ISSUE #3465: Validate the DICL variants from the database instead of config
The programmatic DICL interface gives the following error if the variant already exists in the TOML:

```
2025-09-09T19:30:20.797685Z ERROR tensorzero_core::error: variant 'dicl' already exists in function 'classify_spam' - DICL optimization cannot overwrite existing variants
```

This not correct; we should check if the DICL table in ClickHouse already has any examples for that variant.

anndvision:
closes #3465

### DIFF
--- a/clients/python/tests/test_optimization.py
+++ b/clients/python/tests/test_optimization.py
@@ -13,6 +13,7 @@ from tensorzero import (
     TensorZeroGateway,
     TogetherSFTConfig,
 )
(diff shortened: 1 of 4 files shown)


GabrielBianconi (commented):
LGTM but will let @virajmehta review

virajmehta (approved):

PR #3474, opened 2025-09-10
Add support for `allowed_tools` option in OpenAI-compat

virajmehta:
OpenAI added an [allowed_tools](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice-allowed-tools-allowed_tools) option to its Chat completions endpoint and we did not support it. This PR adds support by refactoring the tool choice handling in the OpenAI compatibility layer. 

We needed a custom deserializer here because the format that they use is not one of `serde`'s formats for enums. 

For now I have only added unit tests for this but if it seems necessary I can also write some integration tests.

### DIFF
--- a/tensorzero-core/src/endpoints/openai_compatible.rs
+++ b/tensorzero-core/src/endpoints/openai_compatible.rs
@@ -382,6 +382,27 @@ struct OpenAICompatibleNamedToolChoice {
     function: FunctionName,
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq)]
+enum OpenAICompatibleAllowedToolsMode {
+    Auto,
+    Required,
+}
+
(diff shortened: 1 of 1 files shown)


tensorzero-core/src/endpoints/openai_compatible.rs
```diff
@@ -382,24 +382,147 @@ struct OpenAICompatibleNamedToolChoice {
...
+                                "Invalid tool choice type: {type_str}",
+                            ))),
+                        }
+                    } else {
+                        Err(D::Error::custom(
+                            "Tool choice 'type' field must be a string",
+                        ))
+                    }
+                } else {
+                    // Object without type field - try to deserialize as OpenAICompatibleAllowedTools directly
```
virajmehta, line 474: we should error here

Aaron1011 (approved):

PR #3527, opened 2025-09-15
[2573] Persist word wrap preference client-side

### LINKED ISSUE #2573: Make `[x] Wrap` button in UI persistent
We should store (client-side) whether the user wants JSON wrapped, and if so, display all JSON fields with that setting (even across pageviews). It can be a cookie or a similar strategy.

BretHudson:
First PR! Whoo!!

I tend to utilize commit messages to break up my PRs into smaller, easy to digest chunks. Kind of pointless for a PR this small, but you get more of a story of how it all came together.

## Description

In order to persist the user's preference for whether or not to wrap JSON, this PR moves the state from React to the browser's local storage.

To accomplish this, I've added a new dependency: `@uidotdev/usehooks` ([usehooks.com](https://usehooks.com/)) to allow us access to their robust `useLocalStorage()` hook.

### Storybook Changes

Storybook needs a decorator added to `preview.tsx` in order to clear browser storage between stories.

## Related

Resolves #2573 

----

> [!IMPORTANT]
> Persist word wrap preference in `CodeEditor` using local storage and update Storybook to reset storage between stories.
> 
>   - **Behavior**:
>     - Persist word wrap preference using `useLocalStorage()` from `@uidotdev/usehooks` in `CodeEditor`.
>     - Default word wrap setting based on language in `code-editor.tsx`.
>   - **Dependencies**:
>     - Add `@uidotdev/usehooks` to `package.json`.
>   - **Storybook**:
>     - Add `resetBrowserStorageDecorator` to `preview.tsx` to clear storage between stories.
>     - Refactor `code-editor.stories.tsx` to use `jsonValue` constant for JSON stories.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 69ab4fea5f1f8fd55897478d83a7a191bff06275. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/components/ui/code-editor.tsx
+++ b/ui/app/components/ui/code-editor.tsx
@@ -154,13 +154,12 @@ export const CodeEditor: React.FC<CodeEditorProps> = ({
     autoDetectLanguage ? detectLanguage(value) : allowedLanguages[0],
   );
 
-  const [_wordWrap, setWordWrap] = useLocalStorage(
+  const [wordWrap, setWordWrap] = useLocalStorage(
     "word-wrap",
-    DEFAULT_WORD_WRAP_LANGUAGES.includes(language).toString(),
+    DEFAULT_WORD_WRAP_LANGUAGES.includes(language),
   );
-  const wordWrap = JSON.parse(_wordWrap);
   const toggleWordWrap = useCallback(() => {
-    setWordWrap((wrap) => (!JSON.parse(wrap)).toString());
+    setWordWrap((wrap) => !wrap);
   }, [setWordWrap]);
   const { copy, didCopy, isCopyAvailable } = useCopy();
   const [mounted, setMounted] = useState(false);
--- a/ui/app/components/ui/code-editor.tsx
+++ b/ui/app/components/ui/code-editor.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { useCopy } from "~/hooks/use-copy";
-import { useLocalStorage } from "@uidotdev/usehooks";
+import { useLocalStorage } from "~/hooks/use-local-storage";
 import CodeMirror from "@uiw/react-codemirror";
 import { json } from "@codemirror/lang-json";
 import { markdown } from "@codemirror/lang-markdown";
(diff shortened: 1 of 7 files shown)


ui/app/components/ui/code-editor.tsx
```diff
@@ -153,7 +154,8 @@ export const CodeEditor: React.FC<CodeEditorProps> = ({
     autoDetectLanguage ? detectLanguage(value) : allowedLanguages[0],
   );
 
-  const [wordWrap, setWordWrap] = useState(
+  const [wordWrap, setWordWrap] = useLocalStorage(
+    "word-wrap",
```
BretHudson, line 158: Since `DEFAULT_WORD_WRAP_LANGUAGES` suggests different languages should have different defaults, would we want the local storage value to be global across all languages, or have it set per-language?

I'm leaning towards it being "global", but I'm not sure I have enough context of how this is being used across the dashboard to make a final decision. Thoughts?

GabrielBianconi: I think let's start with global, don't have an intuition for what would be different from global right now.

virajmehta (changes requested):

ui/app/hooks/use-local-storage.ts
```diff
@@ -0,0 +1,79 @@
...
+  const getSnapshot = () => getLocalStorageItem(key);
+
+  const store =
+    useSyncExternalStore(
+      useLocalStorageSubscribe,
+      getSnapshot,
+      getLocalStorageServerSnapshot,
+    ) ?? "";
+
+  const setState = useCallback<Dispatch<SetStateAction<T>>>(
```
virajmehta, line 52: also generally this file seems pretty complicated for what should be a simple goal. Why's it so complicated?

BretHudson: I agree! React v18 brings `useSyncExternalStore` as the official way to sync external state in React, which adds quite a bit of overhead in needing the define more callback functions.

The code in this file is based off of `@uidotdev/usehooks`'s `useLocalStorage()` hook, which handles plenty of edge cases that can crop up in production.

ui/package.json
```diff
@@ -36,6 +36,7 @@
     "@tailwindcss/vite": "^4.1.11",
     "@tanstack/react-query": "^5.83.0",
     "@tanstack/react-table": "^8.21.3",
+    "@uidotdev/usehooks": "^2.4.1",
```
virajmehta, line 39: do we need this?

virajmehta (approved):

PR #3545, opened 2025-09-16
[1396] Default function now has "Try with model" button

BretHudson:
Draft

### DIFF
--- a/ui/app/components/inference/TryWithButton.tsx
+++ b/ui/app/components/inference/TryWithButton.tsx
@@ -10,15 +10,17 @@ import { ChevronDown } from "lucide-react";
 import { Compare } from "~/components/icons/Icons";
 
 export interface TryWithButtonProps {
-  variants: string[];
-  onVariantSelect: (variant: string) => void;
+  options: string[];
+  onOptionSelect: (variant: string) => void;
   isLoading: boolean;
+  model?: boolean;
 }
 
 export function TryWithButton({
-  variants,
-  onVariantSelect,
+  options: variants,
+  onOptionSelect: onVariantSelect,
   isLoading,
+  model,
 }: TryWithButtonProps) {
   const [isOpen, setIsOpen] = useState(false);
 
@@ -27,7 +29,7 @@ export function TryWithButton({
       <DropdownMenuTrigger asChild>
         <Button variant="outline" size="sm" disabled={isLoading}>
           <ButtonIcon as={Compare} variant="tertiary" />
-          Try with variant
+          Try with {model ? "model" : "variant"}
           <ButtonIcon as={ChevronDown} variant="tertiary" />
         </Button>
       </DropdownMenuTrigger>
--- a/ui/app/components/inference/TryWithButton.tsx
+++ b/ui/app/components/inference/TryWithButton.tsx
@@ -9,17 +9,17 @@ import {
 import { ChevronDown } from "lucide-react";
 import { Compare } from "~/components/icons/Icons";
 
-export interface TryWithVariantButtonProps {
+export interface TryWithButtonProps {
   variants: string[];
   onVariantSelect: (variant: string) => void;
   isLoading: boolean;
 }
 
-export function TryWithVariantButton({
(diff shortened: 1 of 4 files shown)


ui/app/components/inference/TryWithButton.tsx
```diff
@@ -9,25 +9,27 @@ import {
...
 
-export function TryWithVariantButton({
-  variants,
-  onVariantSelect,
+export function TryWithButton({
+  options: variants,
+  onOptionSelect: onVariantSelect,
   isLoading,
-}: TryWithVariantButtonProps) {
+  model,
```
BretHudson, line 23: Using a boolean for now - we could switch this to a string in the future if there are more than the two states

ui/app/routes/datasets/$dataset_name/datapoint/$id/DatapointActions.tsx
```diff
@@ -30,18 +30,18 @@ export function DatapointActions({
...
     <ActionBar>
-      {showTryWithVariant && (
-        <TryWithVariantButton
-          variants={variants}
-          onVariantSelect={onVariantSelect}
+      {showTryWithButton && (
+        <TryWithButton
+          options={variants}
+          onOptionSelect={onVariantSelect}
           isLoading={variantInferenceIsLoading}
```
BretHudson, line 45: Since `showTryWithButton` only gets set to true if the `function_name` is not `"tensorzero::default"`, I don't think we need to do anything here.

ui/app/routes/observability/inferences/$inference_id/route.tsx
```diff
@@ -398,6 +413,23 @@ export default function InferencePage({ loaderData }: Route.ComponentProps) {
...
+
+  const modelsSet = new Set<string>([
+    // models successfully used with default function
+    ...model_inferences.map(({ model_name }) => model_name),
+    // all configured models in config
+    ...Object.keys(config.models),
+    // TODO(bret): list of popular/common model choices
+    // see https://github.com/tensorzero/tensorzero/issues/1396#issuecomment-3286424944
+  ]);
+  const models = [...modelsSet].sort();
```
BretHudson, line 428: Using a Set here to ensure there aren't any duplicates. For now, I thought alphabetical search would be sufficient.

Should the number of options grow to double digits, leveraging a dropdown that has a search field would be reasonable to add here

ui/app/routes/datasets/$dataset_name/datapoint/$id/DatapointActions.tsx
BretHudson: Created a new issue: https://github.com/tensorzero/tensorzero/issues/3569

ui/app/components/inference/TryWithButton.tsx
BretHudson: You've got it!

virajmehta:
One request here: can you please change the test to use a model that is publicly available? Any model with a shorter name is fine bc the `ft:...` models are fine-tuned on our account only. This will make the tests hard to run for an outside developer with an OpenAI key from a different account. One you make that change I can rerun the command.

Otherwise this PR LGTM

ui/e2e_tests/observability.inferences.$inference_id.spec.ts
```diff
@@ -337,61 +337,77 @@ test("should be able to add chat demonstration feedback via the inference page",
...
-    .waitFor({ state: "visible", timeout: 15000 });
-
-  // Click the "Add as Demonstration" button
-  await page.getByText("Add as Demonstration").click();
-
-  // Wait for the modal to close and the page to redirect with newFeedbackId
-  await page.waitForURL((url) => url.searchParams.has("newFeedbackId"), {
-    timeout: 10000,
+test.describe
+  .only("should be able to add demonstration feedback via Try with X flows", () => {
```
BretHudson, line 341: oops, lemme remove that 👀

BretHudson:
Closed in favor of #3574

PR #3573, opened 2025-09-17
[3516] Support default function on variant detail page

### LINKED ISSUE #3516: The variant detail page should support the default function
`/observability/functions/tensorzero::default` → it'll list variants = models used with the default function.

That said, you can't open them. We should add special handling for those variants and actually display a special variant detail page for them.

The list in the function page should include any variants (models) with non-zero usage with this function. I believe this is already the current behavior.

BretHudson:
### Testing

1. Navigate to [/observability/functions/tensorzero::default](http://localhost:5173/observability/functions/tensorzero::default)
2. Click on a variant name
3. Observe page loading

### Related Issues

Resolves #3516

### DIFF
--- a/ui/app/routes/observability/functions/$function_name/route.tsx
+++ b/ui/app/routes/observability/functions/$function_name/route.tsx
@@ -38,6 +38,7 @@ import {
 } from "~/components/layout/PageLayout";
 import { getFunctionTypeIcon } from "~/utils/icon";
 import { logger } from "~/utils/logger";
+import { DEFAULT_FUNCTION } from "~/utils/constants";
 
 export async function loader({ request, params }: Route.LoaderArgs) {
   const { function_name } = params;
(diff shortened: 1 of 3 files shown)


ui/app/routes/observability/functions/$function_name/route.tsx
```diff
@@ -123,6 +124,32 @@ export async function loader({ request, params }: Route.LoaderArgs) {
         weight: 0,
       },
     };
+
+    if (function_name === DEFAULT_FUNCTION) {
+      variant_config = {
+        inner: {
+          type: "chat_completion",
+          model: DEFAULT_FUNCTION,
```
virajmehta, line 132: The model here should be the variant name not DEFAULT_FUNCTION

virajmehta (changes requested):
tiny change to the model name and then can you add a playwright test that this page loads properly. Otherwise, LGTM!

Issue #3376, opened 2025-09-02
Change fields of initialized variant configs to private with getters and remove the `Deserialize` impl


virajmehta:
We have enough validation logic in `load` that we want to force callers to go through that function to get an initialized VariantConfig. We should take ChatCompletionConfig, DiclConfig, etc. and make sure that their fields are private.

dangvu0502:
I’ve opened a PR for this issue. Take a look when you have time

PR #3600, opened 2025-09-20
Make DiclConfig fields private with getters

dangvu0502:
- Removed `#[derive(Deserialize)]` from `DiclConfig` struct
- Added getter methods for all fields
- Added `set_weight()` method (required by the evaluations system)

Related to #3376

After this PR is approved, I will apply the same pattern to the remaining configs in separate PRs

----

> [!IMPORTANT]
> Encapsulate `DiclConfig` fields with private access and public getters, adding a constructor and `set_weight()` method.
> 
>   - **Behavior**:
>     - `DiclConfig` fields made private, with public getter methods added for each field.
>     - `set_weight()` method added to `DiclConfig` for modifying weight.
>     - `new()` constructor added to `DiclConfig` for controlled instantiation.
>   - **Code Adjustments**:
>     - Removed `#[derive(Deserialize)]` from `DiclConfig`.
>     - Updated references to `DiclConfig` fields to use getter methods in `mod.rs`, `evaluations/mod.rs`, and `tests.rs`.
>   - **Misc**:
>     - Related to issue #3376.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for d3aae987737b99e4e0c866ec77a39bf81b25c8f5. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tensorzero-core/src/variant/dicl.rs
+++ b/tensorzero-core/src/variant/dicl.rs
@@ -70,46 +70,6 @@ pub struct DiclConfig {
 }
 
 impl DiclConfig {
-    /// Create a new DiclConfig - this is the only way to construct one from outside the module
-    #[expect(clippy::too_many_arguments)]
-    pub fn new(
-        weight: Option<f64>,
-        embedding_model: String,
-        k: u32,
-        model: String,
-        system_instructions: String,
-        temperature: Option<f32>,
-        top_p: Option<f32>,
-        presence_penalty: Option<f32>,
-        frequency_penalty: Option<f32>,
-        max_tokens: Option<u32>,
-        seed: Option<u32>,
-        json_mode: Option<JsonMode>,
-        extra_body: Option<ExtraBodyConfig>,
-        extra_headers: Option<ExtraHeadersConfig>,
-        retries: RetryConfig,
-        stop_sequences: Option<Vec<String>>,
-    ) -> Self {
-        Self {
-            weight,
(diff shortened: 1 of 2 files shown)


tensorzero-core/src/variant/dicl.rs
```diff
@@ -45,28 +45,138 @@ use super::{
...
     #[cfg_attr(test, ts(skip))]
-    pub extra_headers: Option<ExtraHeadersConfig>,
-    pub retries: RetryConfig,
+    extra_headers: Option<ExtraHeadersConfig>,
+    retries: RetryConfig,
+}
+
+impl DiclConfig {
+    /// Create a new DiclConfig - this is the only way to construct one from outside the module
+    #[expect(clippy::too_many_arguments)]
```
virajmehta, line 74: can we make this `#[cfg(test)]` or remove it if unused?
that would require going through `UninitializedDiclConfig::load()` in the evaluator but I think that's actually good practice and the primary purpose of this refactor (always using load()).

virajmehta (changes requested):
@dangvu0502 thanks for contributing. I think this is almost there but we should remove the impl of `new()` and instead use `load()`. Happy to approve once that change is made.

virajmehta (approved):
LGTM! thanks for the contribution

PR #3714, opened 2025-09-26
Only require max_tokens when token rate limits apply

### LINKED ISSUE #3648: `max_tokens` should not be required if there is no relevant rate limit
```toml
[[rate_limiting.rules]]
priority = 0
model_inferences_per_minute = 1
scope = [
    { tag_key = "application_id", tag_value = "1" }
]
```

```python
from tensorzero import TensorZeroGateway

t0 = TensorZeroGateway.build_http(gateway_url="http://localhost:3000")

response = t0.inference(
    model_name="openai::gpt-5-mini",
    input={
        "messages": [
            {
                "role": "user",
                "content": "Tell me a fun fact.",
            }
        ]
    },
    tags={
        "application_id": "1",
    },
)

print(response)
```

→ 

```
tensorzero.types.TensorZeroError: TensorZeroError (status code 502): {"error":"All variants failed with errors: openai::gpt-5-mini: All model providers failed to infer with errors: openai: Missing max tokens for rate limited request."}
```

---

This works:

```python
from tensorzero import TensorZeroGateway

t0 = TensorZeroGateway.build_http(gateway_url="http://localhost:3000")

response = t0.inference(
    model_name="openai::gpt-5-mini",
    input={
        "messages": [
            {
                "role": "user",
                "content": "Tell me a fun fact.",
            }
        ]
    },
    params={
        "chat_completion": {
            "max_tokens": 1000,
        }
    },
    tags={
        "application_id": "1",
    },
)

print(response)
```

---

**We should only require `max_tokens` if there is a rate limit with tokens that is about to be enforced.**

dangvu0502:
Fixes #3648

### DIFF
--- a/tensorzero-core/src/inference/types/mod.rs
+++ b/tensorzero-core/src/inference/types/mod.rs
@@ -948,6 +948,7 @@ impl RateLimitedRequest for ModelInferenceRequest<'_> {
             extra_headers: _,
             extra_cache_key: _,
         } = self;
+
         let system_tokens = system
             .as_ref()
             .map(|s| get_estimated_tokens(s))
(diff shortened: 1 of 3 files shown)


virajmehta (changes requested):
@dangvu0502 I think this approach is not gonna be very maintainable;

Instead, we might consider changing the `RateLimitedRequest` and `RateLimitedInputContent` to take a `Vec<RateLimitedResource>` and then have it return a new type `EstimatedRateLimitResourceUsage` struct with optional fields. That way if the tokens aren't needed we can set that to None rather than throw an error. Do you mind changing this to that pattern? Thanks and sorry for the trouble here, I know this was underspecified.

dangvu0502:
Got it. I’ll update to follow this pattern

virajmehta:
Hey really sorry Gabriel and I crossed work streams and he ended up closing this issue. Sorry for the trouble here!

PR #3761, opened 2025-09-30
Fix typing for python JsonInferenceOutput

anndvision:


tensorzero-core/src/stored_inference.rs
```diff
@@ -526,7 +526,21 @@ impl RenderedSample {
                         .collect::<PyResult<Vec<_>>>()?;
                     PyList::new(py, output).map(Bound::into_any)
                 }
-                StoredOutput::Json(output) => Ok(output.clone().into_py_any(py)?.into_bound(py)),
+                StoredOutput::Json(output) => {
+                    use crate::inference::types::pyo3_helpers::import_json_inference_output;
```
virajmehta, line 530: this is legitimate actually

tensorzero-core/src/inference/types/mod.rs
```diff
@@ -1192,6 +1192,23 @@ impl std::fmt::Display for JsonInferenceOutput {
...
 #[pymethods]
 impl JsonInferenceOutput {
+    #[new]
+    #[pyo3(signature = (*, raw=None, parsed=None))]
+    fn new(
+        py: Python<'_>,
+        raw: Option<String>,
+        parsed: Option<Bound<'_, PyAny>>,
+    ) -> PyResult<Self> {
+        use pyo3_helpers::deserialize_from_pyobj;
```
virajmehta, line 1202: can we do this in the imports section? the typical way to do it is put
```
#[cfg(feature = "pyo3")]
use pyo3_helpers::deserialize_from_pyobj;
```

virajmehta (changes requested):

PR #3794, opened 2025-10-02
Modify update_datapoint_handler to update the entire datapoint in-place

shuyangli:
Previously `update_datapoint_handler`​  was only serving the "save datapoint" workflow, and wasn't meaningfully differentiated from "add datapoint": when updating datapoints, we always drop episode_id and treat the datapoint as custom. This change supports only partially updating "name" (in a follow-up) while keeping all other data intact.

Note that the `update_datapoint_handler`​ previously removes `episode_id`​, `is_deleted`​, and `staled_at`​ on the backend; we will instead remove them in the frontend for the current "save datapoint" workflow.

These endpoints should probably be redesigned, but we will tackle it as we move ClickHouse queries from Typescript to Rust.


virajmehta:
Agree we need to redesign the API; it was built in a rush and we moved over to other things. I think we should make it more idiomatic CRUD next week or so.

tensorzero-core/src/endpoints/datasets.rs
```diff
@@ -1468,11 +1470,16 @@ where
     }
 }
 
-// SyntheticChatInferenceDatapoints are created by `update_datapoint_handler` and represent datapoints not derived from inferences. In storage their `source_inference_id` is null.
+// UpdateChatInferenceDatapointsReques is used in `update_datapoint_handler` to update an existing ChatInferenceDatapoint
```
virajmehta, line 1473: spelling

tensorzero-core/src/endpoints/datasets.rs
```diff
@@ -1468,11 +1470,16 @@ where
...
+// in place. It's not a partial update - the entire request is written to the database.
+//
+// TODO(shuyangli): this is currently being manually kept in sync with `ChatInferenceDatapoint`. Can we fix this?
```
virajmehta, line 1476: we can do a two-way type assertion like we do [here](https://github.com/tensorzero/tensorzero/blob/d7a5dd59bcfdcf9266d8eed42a9b65f0e78161e8/clients/rust/src/client_inference_params.rs#L139). I think this should be a separate change and maybe subsumed by our larger API refactor but we do have a tool for this and I think it's a good point.

ui/app/routes/datasets/$dataset_name/datapoint/$id/datapointOperations.server.ts
```diff
@@ -40,6 +45,50 @@ function transformOutputForTensorZero(
...
+    name: datapoint.name,
+    tool_params: datapoint.tool_params,
+    staled_at: datapoint.staled_at,
+  };
+  return transformed;
+}
+
+/**
+ * Transforms a JSON datapoint for submission to the TensorZero client.
+ * Generates a new UUID and marks the datapoint as custom.
```
virajmehta, line 72: I think this is incorrect, the ID is generated below.

ui/app/routes/datasets/$dataset_name/datapoint/$id/datapointOperations.server.ts
```diff
@@ -40,6 +45,50 @@ function transformOutputForTensorZero(
   }
 }
 
+/**
+ * Transforms a chat datapoint for submission to the TensorZero client.
+ * Generates a new UUID and marks the datapoint as custom.
```
virajmehta, line 50: I think this is also incorrect.

ui/app/routes/datasets/$dataset_name/datapoint/$id/datapointOperations.server.ts
```diff
@@ -40,6 +45,50 @@ function transformOutputForTensorZero(
...
+ */
+function transformChatDatapoint(datapoint: ParsedChatInferenceDatapointRow): Datapoint {
```
virajmehta, line 52: can we add a more detailed name `transformChatDatapointForUpdate` so it is a little more clear what this is?

virajmehta (changes requested):
Generally looks good and improves code quality but had a couple small issues to clean up. Thanks for the PR!

virajmehta (approved):
LGTM!

virajmehta:
@shuyangli not sure why this failed: https://github.com/tensorzero/tensorzero/actions/runs/18265522285/job/51999280635

but it seems unrelated. We can confirm that and re-enqueue

PR #4122, opened 2025-10-22
Add `allowed_tools` type to ToolCallConfig

virajmehta:
This PR is a no-op until we migrate database and start writing.

Step 1 of our Tools v2 refactor.


tensorzero-core/src/tool.rs
```diff
@@ -163,6 +163,24 @@ pub struct DynamicImplicitToolConfig {
...
+#[cfg_attr(test, ts(export))]
+pub struct AllowedTools {
+    pub tools: Vec<String>,
+    pub choice: AllowedToolsChoice,
+}
+
+#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize, PartialEq, ts_rs::TS)]
+#[cfg_attr(test, ts(export))]
+#[serde(rename_all = "snake_case")]
+pub enum AllowedToolsChoice {
```
shuyangli: Add docstring to what each enum arm means per spec?

tensorzero-core/src/providers/openai/mod.rs
shuyangli: nit: Should we consider `impl Default for ToolCallConfig` so we don't have to awkwardly change this everywhere? (we can do this separately to not mix it into this PR)

shuyangli (dismissed):

tensorzero-core/src/providers/openai/mod.rs
virajmehta: yeah we prob should. I can impl the Default right now and then we can incrementally use it

shuyangli (approved):

Issue #3920, opened 2025-10-13
Allow user to create datapoints from inference query programmatically


GabrielBianconi:
`experimental_list_inferences` + `create_datapoints` in one go

- `create_datapoints_from_inference_query(dataset_name="x", output_source="xxx", ...)` with params from `experimental_list_inferences`
- `POST /datasets/{dataset_name}/datapoints` or `POST /datasets/{dataset_name}/from_inference_query` ?

Issue #3918, opened 2025-10-13
Allow user to create datapoints from inference IDs programmatically


GabrielBianconi:
- `create_datapoints_from_inference_ids(dataset_name="x", inference_ids=[a, b, c], output_source="xyz")`
- `POST /datasets/{dataset_name}/datapoints`

PR #4176, opened 2025-10-24
Support querying inferences by IDs

shuyangli:
A step towards #3918 and #3920.

This PR adds an option to filter by inference IDs in `ListInferencesParams` and adds associated SQL building logic. The newly added "filter by inference" behavior is purely additive and back-compat:
- If only `function_name` is provided, this is essentially a no-op (some of the columns are reordered to make SQL construction easier, but we select the same things - we select both `output_schema` and `tool_params` for both Chat and Json inferences, but when we deserialize the unused one is ignored).
- If both `function_name` and `IDs` are present, the generated SQL is `SELECT ... FROM Chat|JsonInference i WHERE i.function_name = function_name AND i.id in [{ids}]`. This is the same query as above, with an additional ID filter.
- If only `IDs` are present, the generated SQL is `SELECT .. FROM ChatInference i WHERE i.id IN [..] UNION ALL SELECT .. FROM JsonInferences i WHERE i.id IN [..]`. This is the same behavior as Datapoints selection, where we select different kinds of IDs in the same query and deserialize into separate struct types.
  * If only `IDs` are present, and the query contains an ORDER BY metric filter clause, we return an error because there are no guarantees that these IDs are correct functions that have the correct metrics.

We should probably consider relaxing the behavior when only `function_name` is provided and query both tables, because function definitions may not be immutable in practice. Filed #4181 for that.

### DIFF
--- a/tensorzero-core/tests/e2e/list_inferences.rs
+++ b/tensorzero-core/tests/e2e/list_inferences.rs
@@ -1,4 +1,5 @@
 use chrono::DateTime;
+use tensorzero::test_helpers::make_embedded_gateway;
 use tensorzero::{
     BooleanMetricFilter, FloatComparisonOperator, FloatMetricFilter, InferenceFilter,
     InferenceOutputSource, ListInferencesParams, StoredInference, TagComparisonOperator, TagFilter,
@@ -9,7 +10,7 @@ use uuid::Uuid;
 
 #[tokio::test(flavor = "multi_thread")]
 pub async fn test_simple_query_json_function() {
-    let client = tensorzero::test_helpers::make_embedded_gateway().await;
+    let client = make_embedded_gateway().await;
     let order_by = vec![OrderBy {
         term: OrderByTerm::Timestamp,
         direction: OrderDirection::Desc,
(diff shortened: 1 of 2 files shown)


tensorzero-core/src/db/clickhouse/inference_queries.rs
```diff
@@ -67,48 +65,271 @@ pub(crate) fn generate_list_inferences_sql(
...
+
+            // Add all joins (from filters and ORDER BY) before WHERE clause
+            if !joins.get_clauses().is_empty() {
+                if let Some(where_pos) = query.find("\nWHERE") {
+                    query.insert_str(where_pos, &joins.get_clauses().join("\n"));
+                } else {
+                    // No WHERE clause, append at end
+                    query.push_str(&joins.get_clauses().join("\n"));
+                }
+            }
```
virajmehta: this seems quite brittle; not sure what the solution is though; maybe we should change the return type of generate_single_table_query_for_type to return a more structured object and then render it in this function only.

tensorzero-core/tests/e2e/list_inferences.rs
```diff
@@ -427,3 +385,261 @@ async fn test_combined_time_and_tag_filter() {
...
+        order_by: Some(&order_by),
+        ..Default::default()
+    };
+    let res = client.experimental_list_inferences(opts).await;
+
+    // Should error because ORDER BY metric is not supported without function_name
+    assert!(res.is_err());
+    let err_msg = format!("{:?}", res.unwrap_err());
+    assert!(err_msg.contains("not supported"));
+}
```
virajmehta: can you also add a test covering nonexistent IDs

PR #4197, opened 2025-10-26
Fall back to deserializing from dict in deserialize_optimization_config

### LINKED ISSUE #4184: Fall back to deserializing from dict in `deserialize_optimization_config`
We should accept dictionaries as well as our PyO3 types,  this should be an easy fix.

anndvision:
# Description

This PR adds support for passing plain Python dictionaries as optimization configs, in addition to the existing PyO3 class instances.

## Changes

**Rust**:
- Added missing import and check for UninitializedGCPVertexGeminiSFTConfig in deserialize_optimization_config
- Added fallback to deserialize_from_pyobj when the object is not a recognized PyO3 class instance
- Improved error messages to list all supported optimization config types and mention that both class instances and dictionaries are accepted

**Python**:
- Updated OptimizationConfig type alias to include Dict[str, Any] for type checking compatibility
- Modified async optimization tests to use plain dictionaries instead of PyO3 classes
- Kept sync optimization tests using PyO3 classes (tests both deserialization paths)
- Added 4 error handling tests to verify helpful error messages for invalid configs

## Testing

All existing sync tests continue to pass with PyO3 class instances, and async tests verify dictionary-based configs work correctly. Error handling tests ensure users receive clear feedback when passing invalid configurations.

Fixes #4184

----

> [!IMPORTANT]
> Support deserializing optimization configs from Python dictionaries in `deserialize_optimization_config` with improved error handling and updated tests.
> 
>   - **Behavior**:
>     - `deserialize_optimization_config` in `pyo3_helpers.rs` now supports deserializing from Python dictionaries if the object is not a recognized PyO3 class instance.
>     - Improved error messages in `deserialize_optimization_config` to list all supported config types and mention dictionary support.
>   - **Python**:
>     - Updated `OptimizationConfig` type alias in `__init__.py` to include `Dict[str, Any]`.
>     - Modified async tests in `test_optimization.py` to use dictionaries for optimization configs.
>     - Added 4 error handling tests in `test_optimization.py` for invalid config scenarios.
>   - **Misc**:
>     - Added missing import for `UninitializedGCPVertexGeminiSFTConfig` in `pyo3_helpers.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 9c8e66e64ed53dc4df2f36b2a7abdc32be11523d. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


clients/python/tests/test_optimization.py
```diff
@@ -342,13 +338,14 @@ async def test_async_together_sft(
     embedded_async_client: AsyncTensorZeroGateway,
     mixed_rendered_samples: List[RenderedSample],
 ):
-    optimization_config = TogetherSFTConfig(
```
virajmehta: we should leave some of these as native types but maybe just add one that tests the dict path so we have good coverage overall

virajmehta (changes requested):
one small change; otherwise, LGTM

clients/python/tests/test_optimization.py
anndvision: interleaved the dict and native types across sync and async tests for full coverage, rather than sync native and async dict

virajmehta (dismissed):

virajmehta (approved):

PR #4355, opened 2025-11-02
Add initial work towards inference params V2; support `reasoning_effort` and `verbosity` and `thinking_budget_tokens`

GabrielBianconi:
I'm creating a new struct `ChatCompletionInferenceParamsV2` (to be renamed later) which will contain the inference params, and it'll be threaded all the way down. I'll move other params here in subsequent PRs, to keep scope small, then fully merge with `ChatCompletionInferenceParams`.

Once `ChatCompletionInferenceParamsV2` makes all the way to each provider, we'll make sure each provider handles or warns about each param. Right now this is manually, but I might revisit with a trait.

The goal is to have near-100% coverage on the OpenAI parameters, with appropriate handling for providers that do not support them (e.g. warn or best-effort convert).

`thinking_budget_tokens` is a non-standard param that merges the value for anthropic and gemini. It's `i32` because some providers have a special `-1` value.

**MUST NUKE CLICKHOUSE BATCH DATA TO MERGE**

----

> [!IMPORTANT]
> Add `ChatCompletionInferenceParamsV2` to handle new inference parameters and integrate into providers and tests.
> 
>   - **Behavior**:
>     - Introduces `ChatCompletionInferenceParamsV2` for handling `reasoning_effort`, `verbosity`, and `thinking_budget_tokens`.
>     - Integrates new params into providers like `anthropic.rs`, `aws_bedrock.rs`, and `azure.rs`.
>     - Updates tests to check for new parameter handling.
>   - **Documentation**:
>     - Updates `batch-inference.mdx`, `inference.mdx`, and `configuration-reference.mdx` to include new parameters.
>   - **Misc**:
>     - Adds `.gitignore` entry for `provider-proxy/request_cache/`.
>     - Minor updates in `merge-queue.yml` and `AGENTS.md`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for e6e0ae0504fef587f30ad8126444890f1dc57886. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tensorzero-core/tests/e2e/providers/batch.rs
+++ b/tensorzero-core/tests/e2e/providers/batch.rs
@@ -1016,8 +1016,6 @@ pub async fn test_start_inference_params_batch_inference_request_with_provider(
                 "top_p": [0.9],
                 "presence_penalty": [0.1],
                 "frequency_penalty": [0.2],
-                "reasoning_effort": ["low"],
-                "verbosity": ["low"],
             },
             "fake_variant_type": {
                 "temperature": [0.8],


internal/tensorzero-auth/bindings/KeyInfo.ts
```diff
@@ -0,0 +1,3 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+export type KeyInfo = { public_id: string, organization: string, workspace: string, description?: string, created_at: string, disabled_at?: string, };
```
GabrielBianconi: ... why was this missing?

tensorzero-core/src/endpoints/openai_compatible.rs
```diff
@@ -714,19 +718,44 @@ impl Params {
...
-        let inference_params = InferenceParams {
-            chat_completion: chat_completion_inference_params,
+
+        let mut inference_params = openai_compatible_params
+            .tensorzero_params
+            .unwrap_or_default();
+
+        // Override the inference parameters with OpenAI-compatible parameters
+        // TODO (GabrielBianconi): Should we warn if we override parameters that are already set?
+        // Currently: OpenAI-compatible parameters take precedence over TensorZero parameters
```
shuyangli: Yes we should, and I think we should let t0 parameters overwrite the OpenAI-compatible ones, because users will have to specifically pass in T0 parameters and they probably expect these to be used. Doesn't have to be in this PR though.

tensorzero-core/src/inference/types/chat_completion_inference_params/mod.rs
```diff
@@ -0,0 +1,33 @@
+//! TODO (GabrielBianconi):
+//! We are migrating the inference parameters to a struct that must be explicitly handled by every model provider.
+//! To avoid a massive PR, I'll start with a small struct as an extension, and gradually migrate the rest of the parameters.
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize, ts_rs::TS)]
+#[ts(export)]
+pub struct ChatCompletionInferenceParamsV2 {
```
shuyangli: Let's add a comment saying this is the canonical superset of inference params we support that gets translated into each provider's specific API shapes.

Is this specifically aiming for ChatCompletion-like APIs? I don't have an alternative (`InferenceParamsV2` also feels incorrectly broad) but I think the naming is confusing (because this is also getting applied on top of the OpenAI responses API).

shuyangli (approved):
Talked offline. To summarize: status quo today is that there are currently separate model rrovider-specific request types that don't have any direct relationship with each other, and they go out of sync, so for many OpenAI-compatible request params we may drop them silently when sending to the model provider. After this PR, requests get translated into this InferenceParamsV2 struct at the OpenAI-compatible handler, then on the way out to each model provider, we explicitly handle each type and warn if they are not supported.

Issue #4489, opened 2025-11-07
Generate complete Python API types in dataclasses


shuyangli:
Right now the types are a mix of hand-written PyO3 stubs and dataclasses; we should do better and generate the API types into dataclasses, to maintain consistency with Rust.

We should also replace all the PyO3 API types with generated dataclasses.

GabrielBianconi:
→ #4990

PR #4491, opened 2025-11-07
Implement JSON Schema to Python Dataclasses generation

shuyangli:
A step towards #4489. This adds scaffolding (`datamodel-code-generator`​, Python templates for dataclass "null / unset / value" handling, and scripts that trigger it with appropriate configs).

To run the generation step, run `pnpm generate-python-schemas`​.

This PR doesn't mark any existing types for generation yet; see [#4493](https://github.com/tensorzero/tensorzero/pull/4493/files) for generated types.

Some notes:

- For UUIDs, dataclass generate `str` fields, so the generated types are different from the existing UUID types: https://github.com/koxudaxi/datamodel-code-generator/issues/1784. They are identifiers and clients should never construct them, so I think this is probably okay; if we really want to be exactly backcompat, we will need to make some custom deserializers (which we do today, but we may want to remove them?)

----

> [!IMPORTANT]
> Add JSON Schema to Python dataclass generation with support for UUIDs and optional fields, including a script and tests for automation.
> 
>   - **Behavior**:
>     - Adds `generate_schema_types.py` to generate Python dataclasses from JSON schemas.
>     - Introduces `pnpm generate-python-schemas` script in `package.json` to run the generation.
>     - Handles UUIDs as `str` fields in generated dataclasses.
>     - Supports `Option<Option<T>>` fields with `UnsetType` sentinel for distinguishing null and unset values.
>   - **Templates**:
>     - Adds `dataclass.jinja2` for custom dataclass generation.
>     - Adds `generated_types_header.py` for including `UnsetType` sentinel.
>   - **Tests**:
>     - Adds `test_generate_schema_types.py` for testing schema generation functions.
>   - **Dependencies**:
>     - Adds `datamodel-code-generator` to `requirements.txt` and `pyproject.toml` for schema generation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 1d26f71ade4b5ece5c5396eb599408125e8301dd. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/clients/python/generate_schema_types.py
+++ b/clients/python/generate_schema_types.py
@@ -1,13 +1,41 @@
 #!/usr/bin/env python3
 """
-Generate Python dataclasses from Rust JSON schemas.
+Generate Python dataclasses from JSON schemas derived from Rust types.
 
-This script:
+
+For pure value types (mostly used in APIs), we use `schemars` to generate the JSON schemas from Rust types,
+and `datamodel-code-generator` to generate the Python dataclasses from the JSON schemas.
+
+Rust Types (with annotations)
+    ↓ (cargo test export_schema)
+JSON Schemas (in clients/schemas/)
+    ↓ (python generate_schema_types.py)
+Python Dataclasses (in clients/python/tensorzero/generated_types)
+
+
+This script runs both steps of the generation pipeline:
+
 1. Runs cargo tests to generate JSON schemas from Rust types (using #[export_schema])
+    - See `tensorzero-derive` crate for more details on the macro, and https://graham.cool/schemars/ for schemars docs.
 2. Generates Python dataclasses from those JSON schemas
+    - For detailed configuration options, see https://koxudaxi.github.io/datamodel-code-generator/.
 
-Run this script from the clients/python directory:
-    python generate_schema_types.py
+It also sets up customizations to make them work well together:
+
+1.  `schemars` generates one JSON Schema file for each Rust type, bringing in all its dependencies. Naively
+    generating Python dataclasses from these JSON Schemas would cause dependencies to be repeatedly generated, so
+    we merge them into a single schema file before generating Python dataclasses.
+2.  When merging the schema files, some references are recursive
+    (e.g. `InferenceFilter`) that point to the root schema of these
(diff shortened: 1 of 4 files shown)


clients/python/generate_schema_types.py
```diff
@@ -0,0 +1,498 @@
+#!/usr/bin/env python3
+"""
+Generate Python dataclasses from Rust JSON schemas.
+
+This script:
+1. Runs cargo tests to generate JSON schemas from Rust types (using #[export_schema])
+2. Generates Python dataclasses from those JSON schemas
```
virajmehta: can you elaborate on how this works here?

virajmehta (changes requested):

examples/docs/guides/gateway/call-any-llm/requirements.txt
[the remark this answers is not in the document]
shuyangli: I don't actually understand why but CI was complaining that the `uv export` step was failing. Let me try deleting and see what is going on.

clients/python/generate_schema_types.py
shuyangli: Done.

Aaron1011 (dismissed):

clients/python/generate_schema_types.py
```diff
@@ -0,0 +1,471 @@
...
+    - For detailed configuration options, see https://koxudaxi.github.io/datamodel-code-generator/.
+
+It also sets up customizations to make them work well together:
+
+1.  `schemars` generates one JSON Schema file for each Rust type, bringing in all its dependencies. Naively
+    generating Python dataclasses from these JSON Schemas would cause dependencies to be repeatedly generated, so
+    we merge them into a single schema file before generating Python dataclasses.
+2.  To generate the __init__.py file so clients can simply write `from tensorzero.generated_types import X`, we
+    collect all the type names from the schema and export them.
+3.  We generate a custom header file for the generated types to include the `UNSET` sentinel value.
```
virajmehta: add something for rewrite_refs_recursive

shuyangli: Done.

Issue #3898, opened 2025-10-12
Convert `console.error` and `console.warn` to appropriate visual cues


GabrielBianconi:
In https://github.com/tensorzero/tensorzero/pull/3858, we refactored the `logger` and noticed that the UI was using `logger` in the browser too much. We should provide visual cues instead (unless N/A).

PR #4522, opened 2025-11-10
Refactor Toast Implementation

chaance:
This refactor makes a few changes to how toasts are implemented:

- Move toast state into the React tree via a `GlobalToastProvider`. Mutating
  global state outside from outside of the tree without useSyncExternalStore is
  an anti-pattern and can lead to tearing.
- Replace `toast()` with  different toast methods for various use-cases:
	- `toast.info` (generic message without semantics)
	- `toast.success`
	- `toast.warn`
	- `toast.error`
- All toast methods now accept a `log` option to log the toast message
	to the console. This is useful for error toasts where we want to log the error
	for debugging purposes.
- Disallow all direct methods to console to discourage logging without a
  user-facing message.

This PR **does not** update places where we are using `logger` in the UI to show toasts, so there are still places where user-facing errors are not shown. This will be addressed in a follow-up, and we can consider adding a custom lint rule to catch logger references in client-code similar that discourages these. I don't think this PR closes https://github.com/tensorzero/tensorzero/issues/3898 but is a first step in addressing it.

----

> [!IMPORTANT]
> Refactor toast implementation with `GlobalToastProvider`, new toast methods, and logging options, while discouraging direct console logging.
> 
>   - **Behavior**:
>     - Introduces `GlobalToastProvider` to manage toast state within React tree.
>     - Replaces `toast()` with `toast.info`, `toast.success`, `toast.warn`, and `toast.error` methods.
>     - Adds `log` option to toast methods for console logging.
>     - Disallows direct console methods to encourage user-facing messages.
>   - **Components**:
>     - Updates `AddToDatasetButton.tsx`, `DatasetSelector.stories.tsx`, and `VariantEditor.stories.tsx` to use new toast methods.
>     - Removes direct `console.log` calls in `EditableChip.tsx` and other components.
>   - **Context and Hooks**:
>     - Adds `toast-context.tsx` for managing toast actions and state.
>     - Refactors `use-toast.ts` to use new context.
>     - Updates `use-copy.ts` and `use-local-storage.ts` to use new toast methods.
>   - **Providers**:
>     - Adds `global-toast-provider.tsx` to provide toast context.
>   - **Misc**:
>     - Adjusts ESLint rules in `eslint.config.js` to warn on `console` usage.
>     - Updates `logger.ts` to format log messages consistently.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 29ea06b6447758a2d04a86d51ebe8dab352c9e37. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/app/components/dataset/AddToDatasetButton.tsx
+++ b/ui/app/components/dataset/AddToDatasetButton.tsx
@@ -51,12 +51,13 @@ export function AddToDatasetButton({
   useEffect(() => {
     if (fetcher.state === "idle" && fetcher.data) {
       if (fetcher.data.error) {
-        toast.error({
+        const { dismiss } = toast.error({
           title: "Failed to add to dataset",
           description: fetcher.data.error,
         });
+        return () => dismiss({ immediate: true });
       } else if (fetcher.data.redirectTo) {
-        toast.success({
+        const { dismiss } = toast.success({
           title: "New Datapoint",
           description: "A datapoint was created successfully.",
           action: (
@@ -65,8 +66,10 @@ export function AddToDatasetButton({
             </ToastAction>
           ),
         });
+        return () => dismiss({ immediate: true });
       }
     }
+    return;
   }, [fetcher.state, fetcher.data, toast]);
 
   // Helper function to handle dataset selection
--- a/ui/app/context/toast-context.tsx
+++ b/ui/app/context/toast-context.tsx
@@ -24,7 +24,7 @@ export type ToastAction =
 
 export interface ToastEmitResult {
   id: ToasterToast["id"];
-  dismiss: () => void;
+  dismiss: (args?: { immediate?: boolean }) => void;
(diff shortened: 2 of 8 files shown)


ui/app/components/ui/toast.tsx
```diff
@@ -28,6 +28,8 @@ const toastVariants = cva(
     variants: {
       variant: {
         default: "border bg-background text-foreground",
+        // TODO: Implement success variant styles
+        success: "border bg-background text-foreground",
```
chaance: Didn't want to change existing styles arbitrarily but I think it's worth having a separate variant here.

virajmehta (approved):

PR #4583, opened 2025-11-12
Provide a way to disable Postgres even if variable is present

### LINKED ISSUE #4304: Provide a way to disable Postgres even if variable is present
Today we can do `gateway.observability.enabled = false` for CH but there's no parallel for PG.

GabrielBianconi:
Fix #4304


GabrielBianconi:
@codex review

PR #4594, opened 2025-11-13
Remove existing StoredInference PyO3 type

shuyangli:
A step towards #4499


clients/python/tests/test_list_inferences.py
```diff
@@ -572,10 +572,10 @@ async def test_simple_query_chat_function_async(
         output_0 = output[0]
         assert output_0.type == "text"
         # Type narrowing: we know it's a Text block
-        assert isinstance(output_0, Text)
+        assert isinstance(output_0, ContentBlockChatOutputText)
```
virajmehta: do the types here need to be trees? Can we have DAGs instead?

tensorzero-core/src/inference/types/pyo3_helpers.rs
```diff
@@ -390,10 +395,11 @@ pub fn deserialize_from_stored_sample<'a>(
...
                 json_wire,
             ))),
-        }
-    } else {
-        deserialize_from_pyobj(py, obj)
+        };
     }
+
+    // Fall back to generic deserialization
+    deserialize_from_pyobj(py, obj)
```
virajmehta: for my own understanding: we need these fallbacks because there is still a possibility that the user has passed a PyO3 type that we know how to accept and we need to directly extract those before trying the more standard deserialization-based approach

clients/python/tests/test_list_inferences.py
shuyangli: i think technically this is expressing that `ContentBlockChatOutput::Text` is a different "text" type from `StoredInputMessageContent::Text` even though their internal values are both `Text` and they both have a `type` field, so this seems ~correct to me; we could instead have a shared `TaggedText` class that's `{ type: str = "text" } & Text` that might be cleaner today from the developer's perspective, but if things ever change then it becomes annoying to undo?

tensorzero-core/src/inference/types/pyo3_helpers.rs
virajmehta: makes sense to me

virajmehta (approved):

PR #4621, opened 2025-11-14
Support experimental text query for inferences

### LINKED ISSUE #3438: Allow users to search for inferences [UI]
We should support free-form text search on inputs and outputs. The database column for these two should be separate, to allow for searching on only either of them.

shuyangli:
This adds parameters for an experimental text query that does a substring match on inference inputs and outputs. We also support ranking by term frequency (externally called "relevance" to allow for future evolution) as an additional OrderBy option.

This just works in the python library.

Fixes #3438.

### DIFF
--- a/internal/tensorzero-node/lib/bindings/ListInferencesRequest.ts
+++ b/internal/tensorzero-node/lib/bindings/ListInferencesRequest.ts
@@ -48,4 +48,19 @@ export type ListInferencesRequest = {
    * Supports multiple sort criteria (e.g., sort by timestamp then by metric).
    */
   order_by?: Array<OrderBy>;
+  /**
+   * Text query to filter. Case-insensitive substring search over the inferences' input and output.
+   *
+   * THIS FEATURE IS EXPERIMENTAL, and we may change or remove it at any time.
+   * We recommend against depending on this feature for critical use cases.
+   *
+   * Important limitations:
+   * - This requires an exact substring match; we do not tokenize this query string.
+   * - This doesn't search for any content in the template itself.
+   * - Quality is based on term frequency > 0, without any relevance scoring.
+   * - There are no performance guarantees (it's best effort only). Today, with no other
+   *   filters, it will perform a full table scan, which may be extremely slow depending
+   *   on the data volume.
+   */
+  search_query_experimental?: string;
 };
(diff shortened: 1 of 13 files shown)


internal/tensorzero-node/lib/bindings/ListInferencesRequest.ts
```diff
@@ -48,4 +48,18 @@ export type ListInferencesRequest = {
...
+   * We recommend against depending on this feature for critical use cases.
+   *
+   * Important limitations:
+   * - This doesn't search for any content in the template itself.
+   * - Quality is based on term frequency > 0, without any relevance scoring.
+   * - There are no performance guarantees (it's best effort only). Today, with no other
+   *   filters, it will perform a full table scan, which may be extremely slow depending
+   *   on the data volume.
+   */
+  text_query_experimental?: string;
```
GabrielBianconi: `search_query_experimental`?

internal/tensorzero-node/lib/bindings/OrderBy.ts
```diff
@@ -18,4 +18,5 @@ export type OrderBy = {
        */
       name: string;
     }
+  | { by: "term_frequency" }
```
GabrielBianconi: RFC: Just `search_relevance` so we can evolve under the hood however we'd like? Later maybe more granular options.

shuyangli: Done.

tensorzero-core/src/db/clickhouse/inference_queries.rs
```diff
@@ -405,6 +424,29 @@ fn generate_single_table_query_for_type(
...
+            "countSubstringsCaseInsensitiveUTF8(i.input, {text_query_param}) as input_term_frequency"
+        ));
+        select_clauses.push(format!(
+            "countSubstringsCaseInsensitiveUTF8(i.output, {text_query_param}) as output_term_frequency"
+        ));
+        select_clauses.push(
+            "input_term_frequency + output_term_frequency as total_term_frequency".to_string(),
+        );
+
+        where_clauses.push("total_term_frequency > 0".to_string());
```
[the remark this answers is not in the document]
shuyangli: that's just incorrect

PR #4627, opened 2025-11-14
Expose adaptive stopping params for evals in CLI, Python client, node, and UI

amishler:
Exposes previously-added arguments for adaptive stopping of evaluators in eval runs. These enable users to:
- cap the number of datapoint inferences in an eval run to something smaller than the size of the dataset,
- stop individual evaluators early during an eval run if their 95% confidence intervals become narrow enough.

This lets users control the cost-precision tradeoff in eval runs.


amishler:
@codex review

PR #4677, opened 2025-11-17
Add GEPA Optimizer Types and Trait Implementations

anndvision:
This PR adds the type definitions and trait implementations for the GEPA (Genetic Evolution with Pareto Analysis) optimizer. **Note: The actual GEPA algorithm is not yet implemented** - this PR focuses on establishing the API/configuration surface for feedback.

## What's Added

- `GEPAConfig` and `UninitializedGEPAConfig` in `tensorzero-core/src/optimization/gepa.rs`
- `GEPAJobHandle` for tracking optimization results
- Trait implementations (`Optimizer` and `JobHandle`) in `tensorzero-optimizers/src/gepa/mod.rs`
- Integration with existing optimizer dispatchers in `tensorzero-optimizers/src/lib.rs`
- Placeholder implementation that returns dummy results with warning logs

## GEPAConfig - Main Focus for Feedback

The `GEPAConfig` defines the user-facing API for GEPA optimization. We need feedback on whether this configuration provides the right level of control and usability:

### Core Parameters

- **`function_name`** (required): Name of the function being optimized
- **`evaluation_name`** (required): Name of the tensorzero evaluation to score variants
- **`initial_variants`** (optional): List of variant names to seed GEPA with (defaults to all variants for the function)
- **`variant_prefix`** (optional): Prefix for newly generated variant names

### Algorithm Control

- **`batch_size`** (default: 5): Number of training samples to analyze and use for template generation per iteration
- **`max_iterations`** (default: 1): Maximum number of evolution iterations
- **`max_concurrency`** (default: 10): Max concurrent inference calls

### Model Configuration

- **`analysis_model`** (required): Model for analyzing inference results
- **`mutation_model`** (required): Model for generating prompt mutations
- **`max_tokens`** (optional): Max tokens for analysis/mutation model calls

### Context Control

- **`include_inference_input_for_mutation`** (default: true): Whether to include full inference input (not just output) in mutation context
  - WARNING: Can cause context overflow with multi-turn conversations or large batches
  - Provides richer context but increases token usage

### Infrastructure

- **`timeout`** (default: 300s): Client timeout for gateway operations
- **`retries`**: Retry configuration for inference calls
- **`seed`** (optional): Random seed for reproducibility

## Changes Based on Feedback

- Made `analysis_model` and `mutation_model` required parameters (no defaults)
- Changed `max_tokens` to optional with no default value
- Renamed Python stub enum variants from `Gepa` to `GEPA`
- Fixed misleading comment about "lib submodule"
- Simplified code by removing unnecessary `handle_result` variable


clients/python/tensorzero/tensorzero.pyi
```diff
@@ -248,6 +248,7 @@ class OptimizationJobHandle:
     FireworksSFT: Type["OptimizationJobHandle"]
     GCPVertexGeminiSFT: Type["OptimizationJobHandle"]
     TogetherSFT: Type["OptimizationJobHandle"]
+    Gepa: Type["OptimizationJobHandle"]
```
GabrielBianconi: `GEPA`

tensorzero-core/src/optimization/gepa.rs
```diff
@@ -0,0 +1,350 @@
...
+    300
+}
+
+fn default_analysis_model() -> String {
+    "anthropic::claude-sonnet-4-5-20250929".to_string()
+}
+
+fn default_mutation_model() -> String {
+    "anthropic::claude-sonnet-4-5-20250929".to_string()
+}
```
GabrielBianconi: This is going to rot very quickly and it's unclear if they'll have one. I don't think we should have a default. Mandatory to provide.

tensorzero-core/src/optimization/gepa.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    #[serde(default = "default_max_iterations")]
+    pub max_iterations: u32,
+
+    #[serde(default = "default_max_concurrency")]
+    pub max_concurrency: u32,
+
+    pub analysis_model: String,
+
+    pub mutation_model: String,
```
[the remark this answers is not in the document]
GabrielBianconi: @anndvision ^

PR #4699, opened 2025-11-18
Support roundtrip datapoints

### LINKED ISSUE #4697: We should support round-tripping datapoints from get_datapoints to update_datapoints

shuyangli:
This flattens `dynamic_tool_params`​ in UpdateDatapointRequest and allows users to round-trip a datapoint from `get_datapoints`​ or `list_datapoints`​, updates a few fields, and send as-is back to `update_datapoints`​ to update them.

Also skips serializing tool params in get_datapoints to avoid accidentally clearing fields in a round-trip.

Fixes #4697.


tensorzero-core/src/endpoints/datasets/v1/types.rs
```diff
@@ -61,19 +61,105 @@ pub struct UpdateChatDatapointRequest {
...
+        existing_tool_params: Option<DynamicToolParams>,
+    ) -> Option<DynamicToolParams> {
+        if self.allowed_tools.is_none()
+            && self.additional_tools.is_none()
+            && self.tool_choice.is_none()
+            && self.parallel_tool_calls.is_none()
+            && self.provider_tools.is_none()
+        {
+            return None;
+        }
```
virajmehta: are we sure this is correct? what if the `existing_tool_params` is Some() and then all the updates are None?

shuyangli: Restructured the entire function.

tensorzero-core/src/endpoints/datasets/v1/update_datapoints.rs
```diff
@@ -1115,40 +1120,6 @@ mod tests {
...
-            let app_state = create_test_app_state();
-            let fetch_context = create_fetch_context(&app_state.http_client);
-            let dataset_name = "test_dataset";
-            let existing = create_sample_chat_datapoint(dataset_name);
-
-            let update = UpdateChatDatapointRequest {
-                id: existing.id,
-                input: None,
-                output: None,
-                tool_params: Some(None), // Explicitly set to null
```
shuyangli: setting to null doesn't make sense anymore

PR #4777, opened 2025-11-21
Add new lines to end of files

shuyangli:
This keeps happening and I don't think we care that much? I feel like either we get rid of the check in precommit config, or ignore all the .yml and .toml files, or we make everyone's editor automatically add new lines.


GabrielBianconi:
We should add a CI check and reject people's PRs.

PR #4778, opened 2025-11-21
Support searching by input and output in datapoints

shuyangli:
Similar to list_inferences, this adds a `search_query_experimental`​ and supports querying for terms in input and output. Also supports ordering by timestamp and search relevance (term frequency).

At some point we will switch the filter to be `LIKE %term%`​ so it can take advantage of indices, but we will try to understand how users will use this first to see if that's sufficient or if we need to build real relevance-based searching.


tensorzero-core/src/db/clickhouse/dataset_queries.rs
```diff
@@ -667,6 +699,60 @@ impl DatasetQueries for ClickHouseConnectionInfo {
...
+) -> Result<String, Error> {
+    let Some(order_by_vec) = order_by else {
+        return Ok("ORDER BY updated_at DESC, id DESC".to_string());
+    };
+    if order_by_vec.is_empty() {
+        return Ok("ORDER BY updated_at DESC, id DESC".to_string());
+    }
+
+    // Validate that if SearchRelevance is used, search_query_experimental must be present
+    for order_spec in order_by_vec {
```
virajmehta: do we need to validate that there's a single search term?

virajmehta: or rather, do we need to have the interface potentially handle multiple terms? given that it is experimental perhaps not worth doing rn

virajmehta (dismissed):
LGTM!

GabrielBianconi (dismissed):

GabrielBianconi (approved):

PR #4795, opened 2025-11-21
Smaller `mock-inference-server` image

GabrielBianconi:
- Debian → Distroless
- This requires getting rid of `wget` used in health checks, so we set up `grpc-health-probe` (which also does HTTP checks)

(Make CI faster)


GabrielBianconi:
@codex review

PR #4896, opened 2025-11-28
Added write for ConfigSnapshot table only

virajmehta:
- Adds config snapshot writing to ClickHouse 
  - Introduces `UnwrittenConfig` wrapper that holds config + snapshot until written to DB.
  - Creates `StoredConfig` type for backward-compatible serialization of historical snapshots. This will be the "wide type"
  - We hash the re-serialized and sorted `StoredConfig` type for a stable hash

## Of potential concern
I had to add a lot of `Box::pin` on futures involving constructing configs because the config object has gotten larger. Since this is only incurred at config parsing time I don't see it as a severe issue. 
  ## Details

  This PR implements the config snapshot writing functionality that was set up in the base migration PR. When the gateway or embedded
  client starts:

  1. Config is loaded and parsed into an `UnwrittenConfig`
  2. A `ConfigSnapshot` is hashed
  3. When `into_config()` is called, the snapshot is written to ClickHouse
  4. The `Config` retains the hash for correlating with inference requests

  The snapshot storage uses upsert semantics - if the same config hash already exists, `created_at` is preserved while `last_used` is
  updated.

  ### Key types

  - `UnwrittenConfig`: Wrapper holding `Config` + `ConfigSnapshot` before DB write
  - `StoredConfig`: Serialization type for DB storage with explicit `From` impls to catch field drift
  - `ConfigSnapshot`: Contains config TOML, extra templates, and Blake3 hash
  - `SnapshotHash`: Wrapper around the hash as a decimal string (for ClickHouse UInt256)

  ## Tests

  - `cargo test-e2e test_write_config_snapshot` - Tests snapshot write and upsert behavior
  - `cargo test-unit test_prepare_table_for_snapshot` - Tests deterministic TOML sorting

----

> [!IMPORTANT]
> Adds configuration snapshot writing to ClickHouse with new types and tests for stable hash and upsert behavior.
> 
>   - **Behavior**:
>     - Adds `ConfigSnapshot` writing to ClickHouse with upsert semantics in `config/mod.rs`.
>     - Introduces `UnwrittenConfig` to hold config and snapshot until DB write.
>     - Creates `StoredConfig` for backward-compatible serialization of snapshots.
>     - Hashes re-serialized `StoredConfig` for stable hash.
>   - **Types and Functions**:
>     - Adds `write_config_snapshot()` in `config/mod.rs` for writing snapshots to ClickHouse.
>     - Implements `UnwrittenConfig` and `StoredConfig` in `config/unwritten.rs` and `config/stored.rs`.
>     - Updates `Config::load_from_toml()` to return `UnwrittenConfig`.
>   - **Tests**:
>     - Updates tests in `tests/e2e/config.rs` to verify snapshot writing and upsert behavior.
>     - Modifies various test files to use `UnwrittenConfig` and test new snapshot functionality.
>   - **Misc**:
>     - Adds `num-bigint` dependency in `Cargo.toml` for handling large numbers in hashes.
>     - Minor workflow changes in `.github/workflows/merge-queue.yml` to enable feature flag for tests.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 866e89909f64d94ee79ea1612089e420f1c93a26. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


tensorzero-core/src/config/mod.rs
```diff
@@ -1199,6 +1251,142 @@ impl Config {
     }
 }
 
+/// This is for privacy of internal types
+pub mod unwritten_config {
```
shuyangli: this turned out to be a pub mod - is this intentional (i'm not sure you can get around it since config writing happens outside of this module)? consider moving this into `unwritten_config.rs`​ file?

PR #4984, opened 2025-12-04
Load the entire config from the gateway to UI

shuyangli:
Add a backend route `/internal/ui-config` that returns the entire serialized initialized `Config` object from the gateway. This will be a big response object (in our e2e fixtures it's ~5MB) but we will only load it once on UI startup. 

Also adds logic on the UI server side to load this config instead of reading config from disk. This is gated behind an env var `TENSORZERO_FEATURE_FLAG__UI_CONFIG_FROM_GATEWAY`.

Before we take out the feature flag, two things to solve:

- Whenever the gateway restarts, we should invalidate the stored config on the frontend and reload. This is not strictly required but should make it easier to keep the config in sync between UI and gateway.
- The NAPI evaluations runner still reads the config from disk, and we need to remove that and use the gateway-sent version. We should change the evaluation runner logic to not take a config path, but the pieces it needs from the config.

In a follow-up I'll make the `Config` object smaller so we only send back the pieces that the UI needs.

----

> [!IMPORTANT]
> Adds a new backend route to fetch the entire configuration from the gateway and updates the UI to use this configuration if a feature flag is set.
> 
>   - **Backend**:
>     - Adds `/internal/ui-config` route in `internal.rs` to return serialized `Config` object.
>     - Implements `ui_config_handler` in `get_config.rs` to handle the new route.
>   - **UI**:
>     - Updates `loadConfig()` in `index.server.ts` to fetch config from gateway if `TENSORZERO_FEATURE_FLAG__UI_CONFIG_FROM_GATEWAY` is set.
>     - Adds `getTensorZeroClient()` in `get-tensorzero-client.server.ts` to initialize `TensorZeroClient`.
>     - Deprecates `getConfigPath()` in `index.server.ts`.
>   - **Environment**:
>     - Adds `TENSORZERO_FEATURE_FLAG__UI_CONFIG_FROM_GATEWAY` to `Env` in `env.server.ts`.
>     - Updates `getEnv()` to handle the new environment variable.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 58603868c57a0a72ba2ebf2139b179ecd8bac8ca. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


virajmehta:
@shuyangli we should be able to poll the hash of the config from the gateway and if that changes reload the config.

Issue #4990, opened 2025-12-05
Migrate generated Python types to `TypedDict`


GabrielBianconi:
Otherwise, people will have to use the classes rather than dicts to get type checking to work.

PR #4998, opened 2025-12-05
Rename _internal_provider_type to provider_type

shuyangli:
This applies to both `Thought` and `ThoughtChunk`. The field shouldn't be considered internal and it will conflict with our Python types when we use Pydantic.

This is a breaking change; we break serialization backcompat here and only return `provider_type`.

A step towards #4990.

----

> [!IMPORTANT]
> Renames `_internal_provider_type` to `provider_type` in `Thought` and `ThoughtChunk`, affecting serialization and maintaining an alias for database interactions.
> 
>   - **Behavior**:
>     - Renames `_internal_provider_type` to `provider_type` in `Thought` and `ThoughtChunk` classes.
>     - Breaks serialization backcompat by only returning `provider_type`.
>     - Maintains alias `_internal_provider_type` for database interactions in `mod.rs` and `streams.rs`.
>   - **Files Affected**:
>     - `generated_types.py`: Updates `Thought` class.
>     - `mod.rs`: Updates `Thought` struct.
>     - `streams.rs`: Updates `ThoughtChunk` struct.
>   - **Documentation**:
>     - Updates examples in `call-the-openai-responses-api.mdx` to use `provider_type`.
>   - **Tests**:
>     - Updates `inference.test.ts` to reflect changes in `provider_type`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 395e5dc89ee4a05f8ced2993412b89f4ab7e0084. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/clients/python/tensorzero/types.py
+++ b/clients/python/tensorzero/types.py
@@ -153,7 +153,7 @@ class Thought(ContentBlock):
     type: str = "thought"
     signature: Optional[str] = None
     summary: Optional[List["ThoughtSummaryBlock"]] = None
-    _internal_provider_type: Optional[str] = None
+    provider_type: Optional[str] = None
 
 
 @dataclass
@@ -288,7 +288,7 @@ def parse_content_block(block: Dict[str, Any]) -> ContentBlock:
             signature=block.get("signature"),
             summary=summary,
             type=block_type,
-            _internal_provider_type=block.get("_internal_provider_type"),
+            provider_type=block.get("provider_type"),
         )
     elif block_type == "unknown":
         return UnknownContentBlock(
@@ -335,7 +335,7 @@ class ThoughtChunk(ContentBlockChunk):
     signature: Optional[str] = None
     summary_id: Optional[str] = None
     summary_text: Optional[str] = None
-    _internal_provider_type: Optional[str] = None
+    provider_type: Optional[str] = None
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


tensorzero-core/src/inference/types/mod.rs
```diff
@@ -993,32 +993,125 @@ pub enum ThoughtSummaryBlock {
...
-        rename = "_internal_provider_type",
-        skip_serializing_if = "Option::is_none"
-    )]
-    #[ts(optional)]
     pub provider_type: Option<String>,
 }
 
+// Custom deserializer to handle both `provider_type` and legacy `_internal_provider_type` fields,
+// allowing round-tripping of our own serialized output which includes both fields.
+// TODO(#5001): The old field is deprecated in 2025.12.1; we can remove the deserializer in 2026.2+.
```
GabrielBianconi: `_internal_provider_type` is stored in the database so this can never be removed. Might as well use `serde(alias)` here instead of a custom deserializer.

tensorzero-core/src/inference/types/streams.rs
```diff
@@ -51,22 +51,99 @@ pub struct TextChunk {
     pub text: String,
 }
 
-#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq)]
```
GabrielBianconi: Same comments as above.

GabrielBianconi (changes requested):

Issue #5062, opened 2025-12-08
Make UI fully standalone


shuyangli:
TensorZero UI should only require a gateway URL, and API key, (and for the time being a Postgres URL). We should get rid of all other requirements.

NAPI database client code:
- [x] getModelUsageTimeseries (#5124)
- [x] getModelLatencyQuantiles (#5124)
- [x] countDistinctModelsUsed (#5123)
- [x] queryEpisodeTable (#5131)
- [x] queryEpisodeTableBounds (#5131)
- [x] queryFeedbackByTargetId (#5227)
- [x] queryDemonstrationFeedbackByInferenceId (#5340)
- [x] queryFeedbackBoundsByTargetId (#5231)
- [x] getCumulativeFeedbackTimeseries (#5269)
- [x] countFeedbackByTargetId (#5262)
- [x] countRowsForDataset (#5251)
- [x] insertRowsForDataset (#5251)
- [x] countDatasets (#5251)
- [x] countDatapointsForDatasetFunction (#5115)
- [x] getFeedbackByVariant (unused, deleting in #5269)

Typescript code:

**inference.server.ts**
- [x] Migrate `countInferencesForEpisode` to gateway (#5132)
- [x] Migrate `queryInferenceById` to gateway (#5072)
- [x] Migrate `queryModelInferencesByInferenceId` to gateway (#5107)
- [x] Migrate `countInferencesByFunction` to gateway (#5372)

**evaluations.server.ts**
- [x] Migrate `countTotalEvaluationRuns` to gateway (#5120)
- [x] Migrate `getEvaluationRunInfo` to gateway (#5130)
- [x] Migrate `getEvaluationRunInfos` to gateway (#5254)
- [x] Migrate `getEvaluationRunInfosForDatapoint` to gateway (#5258)
- [x] Migrate `getEvaluationResults` to gateway (#5273)
- [x] Migrate `getEvaluationStatistics` to gateway (#5346)
- [x] Migrate `countDatapointsForEvaluation` to gateway (#5135)
- [x] Migrate `searchEvaluationRuns` to gateway (#5136)
- [x] Migrate `getEvaluationsForDatapoint` to gateway (#5371)

**workflow_evaluations.server.ts**
- [x] Migrate `getWorkflowEvaluationRuns` to gateway (#5252)
- [x] Migrate `searchWorkflowEvaluationRuns` to gateway (#5252)
- [x] Migrate `countWorkflowEvaluationRuns` to gateway (#5252)
[...]

PR #5069, opened 2025-12-08
Port countInferencesForFunction and countInferencesForVariant

shuyangli:
This adds a route `GET /internal/functions/{function_name}/inference-stats` which takes an optional `variant_name` query param, and returns the count of inferences for the function for the given variant, and ports the frontend to use it instead of `countInferencesForFunction` and `countInferencesForVariant`.

A step towards #5062.


shuyangli:
@codex review

PR #5099, opened 2025-12-09
Explicitly wait for the inferences response before validating navigation

### LINKED ISSUE #5078: e2e_tests/observability.inferences.filter.spec.ts is flaky
These tests wait for browser navigation, which is async and times out often. We need to find better ways to identify that the correct filters are applied. The issue is that the page doesn't obviously show the applied filters, so we may want to actually display them more prominently so we can write good tests against them.

shuyangli:
Fixes #5078.

----

> [!IMPORTANT]
> Adds `applyFiltersAndWait` helper to `observability.inferences.filter.spec.ts` to ensure inferences response is awaited before validation, improving test reliability.
> 
>   - **Behavior**:
>     - Introduces `applyFiltersAndWait(page: Page)` in `observability.inferences.filter.spec.ts` to wait for inferences response and validate navigation.
>     - Ensures filter sheet is closed and table is loaded with data or empty state message.
>   - **Tests**:
>     - Replaces direct filter application with `applyFiltersAndWait` in tests for function name, variant name, episode ID, search query, advanced tag filter, metric filter, clearing function filter, active filter state, combining filters, and pagination.
>     - Improves test reliability by ensuring page state is updated before assertions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for a9a990127646f9a749fac0df8a50c7d82d0839ef. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/ui/e2e_tests/observability.inferences.filter.spec.ts
+++ b/ui/e2e_tests/observability.inferences.filter.spec.ts
@@ -1,4 +1,33 @@
 import { test, expect } from "@playwright/test";
+import type { Page } from "@playwright/test";
+
+/**
+ * Helper to apply filters and wait for results.
+ * Returns after the filter sheet closes and data reloads.
+ */
+async function applyFiltersAndWait(page: Page) {
+  // Click Apply Filters and wait for navigation to complete
+  const responsePromise = page.waitForResponse(
+    (response) =>
+      response.url().includes("/observability/inferences") &&
+      response.status() === 200,
+  );
+  await page.getByRole("button", { name: "Apply Filters" }).click();
+  await responsePromise;
+
+  // Wait for sheet to close - use Apply Filters button as the indicator
+  await expect(
+    page.getByRole("button", { name: "Apply Filters" }),
+  ).not.toBeVisible();
+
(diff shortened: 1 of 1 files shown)


Aaron1011 (approved):

shuyangli:
For future reference, both Gemini and ChatGPT recommends a pattern that verifies both the server-side action and a user-facing assertion for page hydration:

```typescript
// Trigger the action under test
await page.getByRole('button', { name: 'Submit' }).click();

// Synchronize on the server action
await page.waitForResponse(res =>
  res.url().includes('/actions/login') && res.status() === 200
);

// Synchronize on the redirect’s URL change
await page.waitForURL(/dashboard/);

// Assert something stable
await expect(page.getByRole('heading', { name: 'Welcome' })).toBeVisible();
```

Other recommendations: never wait for 'networkidle' because react router might keep event streams alive, and consider adding known UI elements that we can wait for.

An additional approach: we could stub action responses with Playwright's `route`, but I think we might shoot ourselves here with mocking absent other E2E tests.

Issue #4675, opened 2025-11-17
Add a `load_files=bool` field to endpoints listing/getting datapoints and inferences


GabrielBianconi:
Default false.

If set to true, convert `ObjectStoragePointer` to `ObjectStorageFile` (or `ObjectStorageError`).

PR #5107, opened 2025-12-09
Add internal API to get ModelInference by inference ID

shuyangli:
This removes the `queryModelInferencesByInferenceId` call from the frontend, implements the same call in the gateway, and moves frontend file loading logic into `resolve.server.ts`.

This should really be an external API (since it should be useful for autopilot), but we need to sort out the message type (right now it returns `StoredRequestMessage` which seems incorrect).

A step towards https://github.com/tensorzero/tensorzero/issues/5062


ui/app/utils/resolve.server.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Server-side utilities for resolving file content for inputs and model inferences.
+ *
+ * TODO(shuyangli): find a nice way to avoid duplicating all the types with resolved files; possibly lazily load them.
+ */
```
GabrielBianconi: https://github.com/tensorzero/tensorzero/issues/4675

shuyangli: I'm not 100% convinced by that (loading files at the edge has nice properties) but we can say this is the default solution for now

Aaron1011 (approved):

PR #5167, opened 2025-12-12
Use astral-sh/setup-uv in github actions

shuyangli:
This installs `uv` from the official Github releases so the networking should be more reliable than going outside github. https://github.com/astral-sh/setup-uv It also by default caches dependencies on github runners.


GabrielBianconi (changes requested):
Pls pin to a hash instead of `v7`

virajmehta (approved):

shuyangli:
will investigate the post-uv failure https://github.com/tensorzero/tensorzero/actions/runs/20180792692/job/57940452278

PR #5224, opened 2025-12-16
Add playground auto-evaluation support

simeonlee:
## Summary

Add auto-evaluation capability to the playground, allowing users to run evaluations on inference outputs directly from the playground interface. **This feature is behind a feature flag and disabled by default.**

### Feature Flag

To enable playground evaluations, set the environment variable:
```
VITE_TENSORZERO_FF_PLAYGROUND_EVALS=1
```

### Backend Changes
- Update tensorzero-node to support `datapointIds` for evaluating specific datapoints
- Add `internalDynamicVariantConfig` parameter for evaluating edited (non-builtin) variants  
- Add `runNativeEvaluationStreaming` wrapper in native_client.server.ts

### API
- Add `/api/playground/evaluate` streaming endpoint using NDJSON format
- Configure cache modes: inference uses `on` (read from cache), evaluation also uses cache for determinism

### UI Components
- Add `EvaluationCombobox` component for selecting evaluations with evaluator names annotation
- Add `MetricBadge` component for displaying evaluation results with threshold-based red coloring
- Add `AnimatedEllipsis` component for loading states
- Refactor `Combobox` to support `getItemAnnotation` prop for showing annotations in both dropdown and input

### Playground Integration
- Add `usePlaygroundEvaluation` hook encapsulating evaluation streaming logic
- Integrate evaluation selector into playground UI (behind feature flag)
- Display evaluation metrics below each inference output
- Support both builtin and edited variants
- Auto-run evaluation when inference completes and evaluation is selected
- Reset evaluation when function changes

### MetricBadge Threshold Coloring
- Red background/text when value fails threshold (optimize="max" and value < cutoff, or optimize="min" and value > cutoff)
- For booleans: optimize="max" wants true, optimize="min" wants false
- Reuses `isCutoffFailed` from MetricValue for consistency

### Styling
- Update `muted-foreground` color to be more neutral (remove blue tint)

## Test plan
- [ ] Set `VITE_TENSORZERO_FF_PLAYGROUND_EVALS=1` and verify evaluation selector appears
- [ ] Select an evaluation in playground and verify metrics appear below inference
- [ ] Test with edited variants (should still work)  
- [ ] Test clear button (X) on evaluation selector
- [ ] Verify evaluation re-runs when variant config changes
- [ ] Verify evaluation resets when function changes
- [ ] Test threshold coloring: values below cutoff (for max) or above cutoff (for min) show red

### DIFF
--- a/ui/app/components/metric/MetricBadge.tsx
+++ b/ui/app/components/metric/MetricBadge.tsx
@@ -43,12 +43,12 @@ export function MetricBadge({
       className={clsx(
         baseClasses,
         failsThreshold
-          ? "bg-red-100 text-red-700"
+          ? "bg-red-100 text-red-500"
           : "bg-gray-100 text-gray-900",
       )}
     >
       {label && (
-        <span className={failsThreshold ? "text-red-500" : "text-gray-500"}>
+        <span className={failsThreshold ? "text-red-700" : "text-gray-500"}>
           {label}
         </span>
       )}


ui/app/routes/playground/route.tsx
```diff
@@ -361,6 +367,18 @@ export default function PlaygroundPage({ loaderData }: Route.ComponentProps) {
...
+        <EvaluationCombobox
+          selected={selectedEvaluation}
+          onSelect={(value) =>
+            updateSearchParams({ evaluation: value ?? null })
+          }
+          evaluations={config.evaluations}
+          functionName={functionName}
+          disabled={!functionName || !datasetName || variants.length === 0}
+        />
+      </div>
```
virajmehta: can we feature flag this with `VITE_TENSORZERO_FF_PLAYGROUND_EVALS=1`?

ui/app/routes/playground/DatapointPlaygroundOutput.tsx
```diff
@@ -8,28 +8,206 @@ import {
...
+              ) {
+                setEvalState({
+                  status: "success",
+                  result: {
+                    evaluations: event.evaluations ?? {},
+                    evaluatorErrors: event.evaluator_errors ?? {},
+                  },
+                });
+              } else if (
+                event.type === "error" &&
```
virajmehta: can we sort the evaluations here? so they're displayed in a consistent order across datapoints?

ui/e2e_tests/playground.spec.ts
```diff
@@ -355,3 +355,44 @@ test("playground should work with tool config ID different from display name @cr
     page.getByRole("heading", { name: "Inference Error" }),
   ).toHaveCount(0);
 });
+
+test("playground should run evaluation after inference @credentials", async ({
```
virajmehta: we don't actually need credentials here I believe because all the inferences should be cacheable. Try removing it and we can regenerate fixtures to confirm this .

virajmehta (changes requested):

PR #5241, opened 2025-12-17
Migrate countRowsForDataset and insertRowsForDataset

shuyangli:
- Added two new internal API endpoints:
    - `/internal/inferences/count` - Counts inferences matching filter criteria
    - `/internal/datasets/{dataset_name}/datapoints/insert-inferences` - Inserts matching inferences into a dataset

In this endpoint that inserts matching inferences, we no longer attempts to dedupe, and so we can also remove the `rowsSkipped` param in the frontend.


tensorzero-core/src/db/clickhouse/dataset_queries.rs
```diff
@@ -1100,7 +1125,8 @@ mod tests {
                 NULL as staled_at,
                 id as source_inference_id,
                 false as is_custom,
-                '' AS auxiliary
+                '' AS auxiliary,
+                snapshot_hash
```
shuyangli: I think this may have been broken since when we added `snapshot_hash` but no one noticed. This PR adds some clickhouse test coverage which should help.

PR #5327, opened 2025-12-19
Implement durable step functions for top-k variant selection

amishler:
- Implements `fetch_datapoint_ids_step()` and `process_batch_step()`, which are the two step functions that will be used for the durable execution of the top-k variant identification task, plus associated parameter structs.
- Adds a `new()` function for `MeanBettingConfidenceSequences` which is used in unit tests.
- Edits some of the comments in unit tests to be more informative.
- Stacked on #5323 

----

> [!IMPORTANT]
> Implements durable step functions for top-k variant selection and adds a `new()` function to `MeanBettingConfidenceSequence`.
> 
>   - **Behavior**:
>     - Implements `fetch_datapoint_ids_step()` and `process_batch_step()` in `evaluations/src/topk/mod.rs` for durable execution of top-k variant selection.
>     - `fetch_datapoint_ids_step()` paginates and shuffles datapoint IDs.
>     - `process_batch_step()` processes datapoints, updates confidence sequences, and checks stopping conditions.
>   - **Models**:
>     - Adds `new()` function to `MeanBettingConfidenceSequence` in `evaluations/src/betting_confidence_sequences.rs` for creating sequences with no observations.
>   - **Misc**:
>     - Updates comments in unit tests for clarity.
>     - Stacked on #5323.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 2c820f488b33206cf6e6e3e00031376b04352d8e. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #5339, opened 2025-12-19
Add final components for durable execution of top-k variant selection

amishler:
- Implement `TopKTask`, which runs the top-k variant identification task durably.
- Add integration tests for all four global stopping conditions. Includes new test models and a new test evaluator, because the tests require performance that is non-constant across variants.
- Stacked on #5327


virajmehta (changes requested):
again small nits but great work

virajmehta (approved):

virajmehta (dismissed):

PR #5351, opened 2025-12-20
Add a typescript client for the autopilot API

virajmehta:
This also got tangled with some changes that move authz to its own enum instead of calling it approval. They should be straightforward but if I need to split this I can. 

----

> [!IMPORTANT]
> Add TypeScript client for autopilot API and refactor authorization enums.
> 
>   - **TypeScript Client**:
>     - Adds TypeScript client for autopilot API in `internal/tensorzero-node/lib/bindings/`.
>     - Implements `listAutopilotSessions`, `listAutopilotEvents`, `createAutopilotEvent`, and `streamAutopilotEvents` in `TensorZeroClient` in `tensorzero.ts`.
>   - **Authorization Refactor**:
>     - Renames `ToolCallApproval` to `ToolCallAuthorization` in `lib.rs` and `types.rs`.
>     - Updates related types and enums, including `ToolCallAuthorizationStatus` and `ToolCallDecisionSource`.
>   - **Miscellaneous**:
>     - Updates `.cargo/config.toml` to include `autopilot-client` for TypeScript bindings.
>     - Adds `ts-rs` dependency in `Cargo.lock` and `Cargo.toml`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 644c83097ccebdbf617615b0eb2a33aa6bd23027. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


ui/app/utils/tensorzero/tensorzero.ts
```diff
@@ -1792,6 +1801,130 @@ export class TensorZeroClient {
     return (await response.json()) as GetEvaluationResultsResponse;
   }
 
+  // =========================================================================
+  // Autopilot API Methods
+  // =========================================================================
+
+  /**
+   * Lists autopilot sessions with optional pagination.
+   */
```
shuyangli: I think we should have a separate class for autopilot stuff, one because this file is already too big and we're dealing with a lot of merge conflicts here, and two I think autopilot behavior will likely be different from regular gateway requests down the line. I doubt you need both autopilot and regular gateway methods in the same UI views so it shouldn't introduce additional complexity.

shuyangli (changes requested):
Seems fine but I think we should have a separate TS class for AutopilotClient.

A question for my understanding: do all autopilot calls get routed through gateway (so it's UI => client-side gateway => autopilot-side server)? For auth, UI still only needs a gateway API key, and autopilot API key is configured at the gateway?

This seems correct, but let's think about the relationship with relay. I don't think it makes sense to relay autopilot requests (and requiring the central gateway to hold N long-running streaming handlers for edge gateways).

GabrielBianconi (commented):
blind

PR #5362, opened 2025-12-22
Have the gateway spawn autopilot tools

virajmehta:
When we receive a `ToolCallAuthorization` with an approval from the autopilot API endpoints, we now enqueue a task that runs the tool call and sends the result to the autopilot API.

This PR also contains a change that got lost in merges: Move `SideInfo` and `Output` to the `ToolMetadata` trait instead of the implementation traits (`SimpleTool` and `TaskTool`).


virajmehta:
@codex review

PR #5385, opened 2025-12-23
Add 'gateway.metrics.inference_overhead_histogram_buckets' config

Aaron1011:
This is a list of floating-point values - when set, we report a new metric `tensorzero_inference_latency_overhead_seconds_histogram` using the provided bucket values. The actual latency values recorded as the same as for the ''tensorzero_inference_latency_overhead_seconds' metric, which is unchanged.

----

> [!IMPORTANT]
> Adds configurable histogram buckets for inference latency metrics, enhancing observability with custom bucket support.
> 
>   - **Behavior**:
>     - Adds `gateway.metrics.inference_overhead_histogram_buckets` config to specify custom histogram buckets for `tensorzero_inference_latency_overhead_seconds_histogram`.
>     - If set, records metrics using specified buckets; otherwise, defaults to existing behavior.
>   - **Configuration**:
>     - Updates `MetricsConfig` in `gateway.rs` to include `tensorzero_inference_latency_overhead_seconds_histogram_buckets`.
>     - Validates bucket values to ensure they are non-negative, finite, and in ascending order.
>   - **Metrics**:
>     - Modifies `setup_metrics()` in `observability/mod.rs` to configure histogram buckets if specified.
>     - Updates `OverheadSpanExt` in `overhead_timing.rs` to record histogram metrics conditionally.
>   - **Tests**:
>     - Adds `test_prometheus_metrics_custom_histogram_buckets()` in `prometheus.rs` to verify custom bucket configuration.
>     - Ensures no histogram metrics are reported when not configured in `test_prometheus_metrics_inference_helper()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 39a43bc9c21ca415c30d94ae04bd0f4c5b8f1077. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tensorzero-core/src/observability/mod.rs
+++ b/tensorzero-core/src/observability/mod.rs
@@ -1201,8 +1201,11 @@ pub async fn setup_observability_with_exporter_override<T: SpanExporter + 'stati
 pub fn setup_metrics(metrics_config: Option<&MetricsConfig>) -> Result<PrometheusHandle, Error> {
     let mut builder = PrometheusBuilder::new();
 
-    let buckets =
-        metrics_config.and_then(|config| config.inference_overhead_histogram_buckets.as_ref());
+    let buckets = metrics_config.and_then(|config| {
+        config
+            .tensorzero_inference_latency_overhead_seconds_histogram_buckets
+            .as_ref()
+    });
 
     if let Some(buckets) = buckets {
         use metrics_exporter_prometheus::Matcher;
--- a/gateway/tests/prometheus.rs
+++ b/gateway/tests/prometheus.rs
@@ -140,7 +140,7 @@ async fn test_prometheus_metrics_custom_histogram_buckets() {
 observability.enabled = false
 
 [gateway.metrics]
-inference_overhead_histogram_buckets = [0.0001, 1.0, 10]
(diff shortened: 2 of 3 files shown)


tensorzero-core/src/observability/mod.rs
```diff
@@ -1197,8 +1198,26 @@ pub async fn setup_observability_with_exporter_override<T: SpanExporter + 'stati
...
+            .set_buckets_for_metric(
+                Matcher::Full("tensorzero_inference_latency_overhead_seconds".to_string()),
+                buckets,
+            )
+            .map_err(|e| {
+                Error::new(ErrorDetails::Observability {
+                    message: format!("Failed to set histogram buckets: {e}"),
+                })
+            })?;
+    }
```
GabrielBianconi: Let's make this a separate metric rather than overloading (i.e. report both if buckets are configured):

```
tensorzero_inference_latency_overhead_seconds_histogram
```

Overloading is tricky because a fleet of gateways might brick observability, etc. This way we always have one and optionally have the other.

GabrielBianconi (changes requested):

PR #5479, opened 2025-12-31
Add `run_evaluation` autopilot tool

amishler:
- Wraps `run_evaluation_core_streaming()` from the `evaluations` package, allowing autopilot to run evaluations for individual variants.
- Currently there's no http endpoint for running evaluations, but when that's available we'll refactor this to wrap that instead. This currently throws an error if called from an http client.
- This supports named variants only, not dynamic variants.

----

> [!IMPORTANT]
> Add `RunEvaluationTool` to autopilot for running evaluations on datasets, supporting only embedded gateway mode and named variants.
> 
>   - **Behavior**:
>     - Adds `RunEvaluationTool` in `run_evaluation.rs` to run evaluations on datasets or specific datapoints using `run_evaluation_core_streaming()`.
>     - Supports named variants only, not dynamic variants.
>     - Throws error if called from HTTP client, only supports embedded gateway mode.
>   - **Integration**:
>     - Registers `RunEvaluationTool` in `lib.rs` and `mod.rs`.
>     - Implements `run_evaluation()` in `client_ext.rs` and `embedded.rs` for `TensorZeroClient`.
>   - **Testing**:
>     - Adds integration tests in `run_evaluation_tool.rs` to verify tool execution with dataset names, datapoint IDs, precision targets, and error handling.
>     - Mocks `run_evaluation()` in `common/mod.rs` for testing purposes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for e7a25cfc0134aec814bf157aa95728e91945cb2b. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


internal/durable-tools/src/tensorzero_client/client_ext.rs
```diff
@@ -423,4 +423,13 @@ impl TensorZeroClient for Client {
...
         }
     }
+
+    async fn run_evaluation(
+        &self,
+        _params: super::RunEvaluationParams,
+    ) -> Result<super::RunEvaluationResponse, TensorZeroClientError> {
+        Err(TensorZeroClientError::NotSupported(
+            "run_evaluation is only supported via EmbeddedClient".to_string(),
+        ))
```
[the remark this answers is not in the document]
amishler: Fixed this oversight

internal/durable-tools/src/tensorzero_client/client_ext.rs
```diff
@@ -423,4 +432,131 @@ impl TensorZeroClient for Client {
...
+
+                while let Some(update) = receiver.recv().await {
+                    match update {
+                        EvaluationUpdate::RunInfo(_) => {
+                            // Skip RunInfo
+                            continue;
+                        }
+                        update => {
+                            // Ignore write errors to the dummy sink
+                            let _ = evaluation_stats.push(update, &mut dummy_writer);
```
virajmehta: should we log these? would they get logged already?

virajmehta: even if we don't fail I think we should know

anndvision (approved):
LGTM

internal/durable-tools/src/tensorzero_client/client_ext.rs
amishler: They get logged, because in that branch `update` can be either `EvaluationUpdate::Success(evaluation_info)` or `EvaluationUpdate::Error(evaluation_info)`. It's just the serialized output that gets discarded, and actually that comment is misleading because the sink always successfully consumes data: https://doc.rust-lang.org/stable/std/io/fn.sink.html

So I think this is correct, I just changed the comments above to make it clearer.

PR #5573, opened 2026-01-08
Making betting confidence sequences take running intersection of intervals

### LINKED ISSUE #5572: Make betting-based confidence sequences take running intersection of previous intervals
In `betting_confidence_sequences.rs`, intervals should always be intersected with the previous interval so that intervals are nested over time. The current form yields valid confidence sequences, but running intersections can only result in tighter intervals, which are preferable.

amishler:
Closes #5572.

----

> [!IMPORTANT]
> Modify `update_betting_cs` to ensure nested confidence intervals by taking running intersections, with a new test to verify this behavior.
> 
>   - **Behavior**:
>     - Modify `update_betting_cs` in `betting_confidence_sequences.rs` to ensure confidence intervals are nested by taking the running intersection of intervals.
>     - If intersection results in an empty set, return `[mean_est, mean_est]`.
>   - **Tests**:
>     - Add `test_intervals_are_nested_over_updates` to verify that each update produces an interval contained within or equal to the previous interval.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 7327cf82e9f76715c82721ac59d7a83cfa221757. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/evaluations/src/betting_confidence_sequences.rs
+++ b/evaluations/src/betting_confidence_sequences.rs
@@ -393,8 +393,9 @@ pub fn update_betting_cs(
         (m_values[n_grid - 1], n_grid - 1)
     };
 
-    // Intersect with previous bounds to ensure nested confidence sequences.
-    // If the intersection would be empty, ensure the interval still contains mean_est.
+    // Intersect with previous bounds to ensure nested confidence sequences,
+    // unless that would result in an empty set, in which case just return
+    // [mean_est, mean_est].
     let cs_lower = cs_lower_new.max(prev_results.cs_lower).min(mean_est);
     let cs_upper = cs_upper_new.min(prev_results.cs_upper).max(mean_est);
 


evaluations/src/betting_confidence_sequences.rs
```diff
@@ -377,22 +377,27 @@ pub fn update_betting_cs(
...
     } else {
         // Right endpoint is inside confidence set
         (m_values[n_grid - 1], n_grid - 1)
     };
 
+    // Intersect with previous bounds to ensure nested confidence sequences.
+    // If the intersection would be empty, ensure the interval still contains mean_est.
+    let cs_lower = cs_lower_new.max(prev_results.cs_lower).min(mean_est);
+    let cs_upper = cs_upper_new.min(prev_results.cs_upper).max(mean_est);
+
```
anndvision: If the raw interval and previous interval don't overlap, collapsing to contain mean_est preserves validity (the true mean is still covered with high probability) while maintaining the nested property. Is that correct?

anndvision (approved):

evaluations/src/betting_confidence_sequences.rs
amishler: Validity is preserved, but the logic is actually: ensure nested intervals unless this would result in the empty set, in which case return `[mean_est, mean_est]` instead. Changed the docstring to clarify.

anndvision (approved):

PR #5626, opened 2026-01-13
Unify `StaticJSONSchema` and `DynamicJSONSchema`

### LINKED ISSUE #5016: Unify `StaticJSONSchema` and `DynamicJSONSchema`
There isn't really a need for the distinction, and we will need to support `Deserialize` for function configs soon because we need to pass it from Node to Rust via NAPI. We should stop treating them as separate, and also compile the validator when it's used (or may not even need it at all, since Aaron mentioned this wasn't benchmarked).

shuyangli:
Fixes #5016.

The unified JSONSchema type uses lazy async compilation and provides a consistent API for schema validation.

The validation behavior remains the same - schemas loaded from config are validated for correctness at load time, but the compiled validator is only created when first needed.

----

> [!IMPORTANT]
> Unifies `StaticJSONSchema` and `DynamicJSONSchema` into `JSONSchema` with lazy async compilation, maintaining validation behavior and updating tests accordingly.
> 
>   - **Behavior**:
>     - Unifies `StaticJSONSchema` and `DynamicJSONSchema` into `JSONSchema` with lazy async compilation.
>     - Validation behavior unchanged: schemas validated at load time, compiled validator created on first use.
>   - **Code Changes**:
>     - Replaces `StaticJSONSchema` and `DynamicJSONSchema` with `JSONSchema` in `ChatTemplates.ts`, `DynamicImplicitToolConfig.ts`, and `JSONSchema.ts`.
>     - Updates `validate_input()` in `function_config.rs` to be async.
>   - **Tests**:
>     - Updates tests in `dicl.rs`, `analyze.rs`, and `mutate.rs` to use `JSONSchema`.
>     - Ensures tests reflect unified schema handling and async validation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=tensorzero%2Ftensorzero&utm_source=github&utm_medium=referral)<sup> for 51d408241849144a4a1360be8a2444f5b6d256f0. You can [customize](https://app.ellipsis.dev/tensorzero/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tensorzero-core/src/jsonschema_util.rs
+++ b/tensorzero-core/src/jsonschema_util.rs
@@ -32,26 +32,25 @@ impl PartialEq for JSONSchema {
 }
 
 impl Default for JSONSchema {
-    #[expect(clippy::expect_used)]
     fn default() -> Self {
         // Create an empty schema that accepts any object
-        let value = serde_json::json!({});
-        let validator = jsonschema::validator_for(&value)
-            .expect("Empty schema should always compile successfully");
-        Self::new_with_compiled(value, validator)
+        Self {
+            value: serde_json::json!({}),
+            compiled: Arc::new(OnceCell::new()),
+        }
     }
 }
 
 impl JSONSchema {
     /// Creates a new JSONSchema with a pre-compiled validator.
-    #[expect(clippy::expect_used)]
-    fn new_with_compiled(value: Value, validator: Validator) -> Self {
+    fn new_with_compiled(value: Value, validator: Validator) -> Result<Self, Error> {
         let compiled = Arc::new(OnceCell::new());
-        // This cannot fail since the OnceCell is empty
-        compiled
-            .set(validator)
-            .expect("OnceCell should be empty when created");
-        Self { value, compiled }
+        compiled.set(validator).map_err(|e| {
+            Error::new(ErrorDetails::JsonSchema {
+                message: format!("Failed to set validator in OnceCell: {e}"),
+            })
+        })?;
+        Ok(Self { value, compiled })
     }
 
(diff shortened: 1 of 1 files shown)


tensorzero-core/src/endpoints/batch_inference.rs
```diff
@@ -171,18 +171,14 @@ pub async fn start_batch_inference(
...
-        .iter()
-        .enumerate()
-        .try_for_each(|(i, input)| {
-            function.validate_input(input).map_err(|e| {
-                Error::new(ErrorDetails::BatchInputValidation {
-                    index: i,
-                    message: e.to_string(),
-                })
+    for (i, input) in params.inputs.iter().enumerate() {
+        function.validate_input(input).await.map_err(|e| {
```
shuyangli: Should be able to validate all inputs in parallel instead of sequentially (but P2 since it's batch and not latency-sensitive).

tensorzero-core/src/jsonschema_util.rs
```diff
@@ -10,65 +9,72 @@ use crate::config::path::ResolvedTomlPathData;
...
 }
 
-impl Default for StaticJSONSchema {
+impl Default for JSONSchema {
     fn default() -> Self {
+        Self::new_lazy(serde_json::json!({}))
+    }
+}
+
+impl JSONSchema {
```
shuyangli: The current API is confusing. This should have two main ways to build a new schema:

- compile_async(schema): compiles the schema in the background, stores in OnceCell, consumer reads or blocks on the compilation
- compile(schema): compiles inline on first use, blocks on first compile

PR #5656, opened 2026-01-15
Add React Router streaming to inference detail page

simeonlee:
## Summary

Implements streaming for the inference detail page using React Router's deferred data pattern (Suspense/Await) for instant navigation.

## Status: Needs Rebase

This PR's branch is stale and has type conflicts with current main. The streaming pattern is correct but needs to be rebased and updated to match current types.

## Changes
- Loader returns promise instead of awaited values
- Content wrapped in `<Suspense>` with `<InferenceSkeleton>` fallback
- `<Await errorElement={...}>` for component-level error handling
- Uses primitives from #5797

## Dependencies
- #5797 (streaming primitives)

## Test Plan
- [ ] Navigate to inference detail page - shows skeleton instantly
- [ ] Content appears when data resolves
- [ ] Error state shows inference ID in header
- [ ] Feedback pagination works
- [ ] "Try with variant" works
- [ ] "Add to dataset" works

### DIFF
--- a/ui/app/components/ui/error/ErrorContentPrimitives.tsx
+++ b/ui/app/components/ui/error/ErrorContentPrimitives.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useAsyncError } from "react-router";
+import { useAsyncError, isRouteErrorResponse } from "react-router";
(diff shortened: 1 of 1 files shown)


simeonlee:
@codex review

Issue #5691, opened 2026-01-16
Support Postgres-only TensorZero deployments


shuyangli:
See Notion for details.

PR #5692, opened 2026-01-16
Implement chat json inference tables and count queries for Postgres

### LINKED ISSUE #5759: Add partitioning and TTL for Postgres integration

shuyangli:
First step towards #5691. This adds Chat and Json inference tables, and implements the first InferenceCountQueries for Postgres.

Fixes #5759.

Gateway integration is gated behind the `ENABLE_POSTGRES_READ` feature flag.

We have e2e database tests for these, but not endpoint tests yet. Those are coming in a follow-up PR with some additional CI setup.

Key things to note:
- chat_ and json_inferences tables are partitioned daily (to support TTL at the day granularity for large deployments without needing manual database managements; we create partition tables for the next 7 days, and also add a _default partition for any potential backfills.
- it's difficult to convert a regular table to a partitioned table, and we've decided that TTL is important for postgres, so we'll do it now. for smaller deployments, daily partitions are not hard to manage yet, and we could also sweep older data into _default if we decide it's useful.
- we schedule pgcron if available during the migration; it reads `inference_retention_days` from a table `tensorzero_retention_config` (written or cleared by gateway on startup so it can be loaded from config) to decide how to truncate old partitions.
    - this does not truncate default partition; we'll think about it during backfill discussions
- if no retention config is set, then the scheduled function that drops partitions doesn't do anything.
- removed the gin index on input/output; will maybe index a separate expression for full-text search
- Also, unfortunately sqlx doesn't provide mocks and it uses Structs instead of Traits so we won't have good unit tests that assert on the SQL directly. We'll rely mostly on e2e postgres tests.

Immediate follow-up work:
- Added postgres e2e tests but not endpoint e2e tests yet. This is being designed and will come in #5800.
- Will need to separate out clickhouse and postgres tests so we don't unnecessarily run postgres tests across the clickhouse matrix. This is getting merged ahead of this PR in #5808.

See https://github.com/tensorzero/agentic-design-docs/pull/6 for the design.

### DIFF
--- a/tensorzero-core/tests/e2e/docker-compose.live.yml
+++ b/tensorzero-core/tests/e2e/docker-compose.live.yml
@@ -161,7 +161,7 @@ services:
         condition: service_healthy
       gateway-postgres-migrations:
         condition: service_healthy
-    command: [ "bash", "-c", "cd /fixtures && ./load_fixtures_postgres.sh && touch /load_complete_postgres.marker && sleep infinity" ]
+    command: [ "bash", "-c", "cd /fixtures && ./load_fixtures_postgres.sh && touch /load_complete_postgres.marker" ]
     healthcheck:
       test: [ "CMD", "test", "-f", "/load_complete_postgres.marker" ]
       interval: 5s


tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+-- chat_inferences (partitioned by day)
+CREATE TABLE chat_inferences (
+    id UUID NOT NULL,
+    function_name TEXT NOT NULL,
+    variant_name TEXT NOT NULL,
+    episode_id UUID NOT NULL,
+    input TEXT NOT NULL,
+    output TEXT NOT NULL,
+    tool_params TEXT NOT NULL,
+    inference_params TEXT NOT NULL,
```
virajmehta: jsonb?

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+    processing_time_ms INTEGER,
+    ttft_ms INTEGER,
+    tags JSONB DEFAULT '{}',
+    extra_body TEXT,
```
virajmehta: jsonb

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+    dynamic_tools TEXT[],
```
virajmehta: jsonB I think (the tools are an array of objects)

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+    dynamic_provider_tools TEXT[],
+    allowed_tools TEXT,
+    tool_choice TEXT,
+    parallel_tool_calls BOOLEAN,
```
virajmehta: `parallel_tool_calls` truly needs to be optional

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+-- json_inferences (partitioned by day)
+CREATE TABLE json_inferences (
+    id UUID NOT NULL,
+    function_name TEXT NOT NULL,
+    variant_name TEXT NOT NULL,
+    episode_id UUID NOT NULL,
+    input TEXT NOT NULL,
+    output TEXT NOT NULL,
+    output_schema TEXT NOT NULL,
+    inference_params TEXT NOT NULL,
```
virajmehta: these 4 jsonb

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,119 @@
...
+    processing_time_ms INTEGER,
+    ttft_ms INTEGER,
+    tags JSONB DEFAULT '{}',
+    extra_body TEXT,
+    auxiliary_content TEXT NOT NULL,
```
virajmehta: both JSONB Not null

tensorzero-core/src/db/postgres/inference_count.rs
```diff
@@ -0,0 +1,304 @@
...
+            }
+        })
+        .collect())
+}
+
+fn map_sqlx_error(e: sqlx::Error) -> Error {
+    Error::new(ErrorDetails::PostgresQuery {
+        message: e.to_string(),
+    })
+}
```
virajmehta: we should just impl From<sqlx::Error> in error.rs for our error type if we haven't already. Then we can just use `?` operator.

tensorzero-core/tests/e2e/postgres/inference_count_queries.rs
virajmehta: could we write this as e2e tests that could run against either data store? I would much prefer to be able to assert that they work the same

AGENTS.md
```diff
@@ -26,6 +26,14 @@
 - Business logic layer will generate all data that TensorZero is responsible for (e.g. UUIDs for new datapoints, `staled_at` timestamps).
 - Database layer (ClickHouse and/or Postgres) will insert data as-is into the backing database, with the only exception of `updated_at` timestamps which we insert by calling native functions in the database.
 
+## For Postgres (sqlx)
+
+- **Do not use `format!` for SQL queries.** Use `sqlx::QueryBuilder` for dynamic queries.
+  - Use `.push()` for trusted SQL fragments (table names, SQL keywords).
+  - Use `.push_bind()` for user-provided values (prevents SQL injection, handles types).
+  - Use `.build_query_scalar()` for scalar results, `.build()` for row results.
+- For static queries, use `sqlx::query!` or `sqlx::query()` directly.
```
virajmehta: macros are preferred

virajmehta (changes requested):
see comments

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+    PRIMARY KEY (id, created_at)
+) PARTITION BY RANGE (created_at);
+
+-- Indexes for json_inferences
+CREATE INDEX idx_json_inferences_function_variant_episode
+    ON tensorzero.json_inferences(function_name, variant_name, episode_id);
+CREATE INDEX idx_json_inferences_episode_id ON tensorzero.json_inferences(episode_id);
+CREATE INDEX idx_json_inferences_function_id ON tensorzero.json_inferences(function_name, id);
+CREATE INDEX idx_json_inferences_tags ON tensorzero.json_inferences USING GIN (tags);
+
```
virajmehta: both JSONB Not null

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (id, created_at)
+) PARTITION BY RANGE (created_at);
+
+-- Indexes for chat_inferences
+CREATE INDEX idx_chat_inferences_function_variant_episode
+    ON tensorzero.chat_inferences(function_name, variant_name, episode_id);
+CREATE INDEX idx_chat_inferences_episode_id ON tensorzero.chat_inferences(episode_id);
+CREATE INDEX idx_chat_inferences_function_id ON tensorzero.chat_inferences(function_name, id);
+CREATE INDEX idx_chat_inferences_tags ON tensorzero.chat_inferences USING GIN (tags);
```
virajmehta: all of these tool ones + extra_body can be mandatory I believe. the types are mandatory in app layer.

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+    processing_time_ms INTEGER,
+    ttft_ms INTEGER,
+    tags JSONB NOT NULL DEFAULT '{}',
+    extra_body JSONB NOT NULL DEFAULT '[]',
+    dynamic_tools JSONB NOT NULL DEFAULT '[]',
+    dynamic_provider_tools JSONB NOT NULL DEFAULT '[]',
+    allowed_tools JSONB,
+    tool_choice TEXT,
+    parallel_tool_calls BOOLEAN,
+    snapshot_hash BYTEA,
```
virajmehta: these are jsons too, should we be using JSONB?

tensorzero-core/src/db/postgres/migrations/20260116194140_inference_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+    function_name TEXT NOT NULL,
+    variant_name TEXT NOT NULL,
+    episode_id UUID NOT NULL,
+    input JSONB NOT NULL,
+    output JSONB NOT NULL,
+    tool_params JSONB NOT NULL,
+    inference_params JSONB NOT NULL,
+    processing_time_ms INTEGER,
+    ttft_ms INTEGER,
+    tags JSONB NOT NULL DEFAULT '{}',
```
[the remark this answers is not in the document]
shuyangli: I think this is simpler; we can also index JSONB with GIN and if our queries are all "this key matches this value" then GIN should work well enough; plus with TTL I think we have to reason about cascading truncates.which could get into tricky territories if we don't partition the tags table. I'd vote do this as-is and migrate down the line if we find a compelling reason to have a separate table.

PR #5720, opened 2026-01-19
Fix issues with Alert Dialog and toasts

chaance:
First I noticed that the DatasetTable was incorrectly using `Dialog` instead of `AlertDialog` (which is appropriate because this is a destructive action). While replacing it I also noticed there were 2 alert dialog components, but one was actually replicating the behavior of toasts. This PR fixes both issues and removes some other dead code I found looking for references to the deleted alert dialog component.

### DIFF
--- a/ui/e2e_tests/datasets.datapoint.try-with.spec.ts
+++ b/ui/e2e_tests/datasets.datapoint.try-with.spec.ts
@@ -71,7 +71,7 @@ test.describe("Try with on datapoint page", () => {
       await deleteButton.click();
 
       // Wait for the dialog to appear and click the Delete button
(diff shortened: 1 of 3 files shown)


ui/app/components/ui/TableItems.tsx
```diff
@@ -84,49 +84,53 @@ function TableItemFunction({
...
+                  description:
+                    "This function is not present in your configuration file.",
+                });
+              }}
+            >
+              {content}
+            </button>
+          )}
+        </TooltipTrigger>
+        <TooltipContent align="start">
```
chaance: Removed the override styles from the content here so that this appears the same as other toasts

PR #5726, opened 2026-01-20
fix arguments in haiku hidden preferences example

ndoherty-xyz:
While working through the **haiku-hidden-preferences** example I ran into this error in the Jupyter notebook

```
Error occurred: TensorZeroError: TensorZeroError (status code 400): {"error":"input.messages[0].content[0]: Unknown key `value` in text content"}
```

After a bit of investigation I found that `internal/tensorzero-types/src/message.rs` had the key as `arguments` instead of `value`. Changing that let me work through the rest of the example.


ndoherty-xyz:

virajmehta (dismissed):

virajmehta (approved):

GabrielBianconi:
Good catch @ndoherty-xyz - we deprecated that syntax but looks like we missed this instance. Thanks!

virajmehta (approved):

PR #5747, opened 2026-01-20
Add 'response_format' parameter to list_datapoints

Aaron1011:
This allows specifying `"response_format": "id"`, which will only return a single `"ids"` array containing all of the datapoint ids. All of the query parameters are identical (and we use the same ClickHouse query for now)

For now, this is unsupported in the Python client, and we throw an error if that format is requested

---

> [!NOTE]
> Introduces selectable response shape for dataset listing.
> 
> - Adds `DatapointResponseFormat` and `ListDatapointsResponse` (union) to datasets v1; `list_datapoints` handler now returns either `Datapoints` or `IdsOnly` using the same underlying query
> - Updates Rust client API to return `ListDatapointsResponse`; adjusts implementations, tests, and call sites (errors when `IdsOnly` appears where full datapoints are expected)
> - Extends TS bindings with `DatapointResponseFormat`, `ListDatapointsResponse`, and `ListDatapointsIdsOnlyResponse`
> - Updates internal tools (`autopilot-tools`, `durable-tools`, worker) interfaces and mocks to the new response type
> - Modifies evaluations code/tests to handle the new enum response
> - Python client surfaces the new response but explicitly errors when `response_format: id` is requested in current methods
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1ab94b4efd69b980a154c3eddd2a4a78d5e082f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


clients/python/src/lib.rs
```diff
@@ -2359,9 +2374,25 @@ impl AsyncTensorZeroGateway {
...
+                    ListDatapointsResponse::Datapoints(dp) => {
+                        Ok(PyList::new(py, dp.datapoints)?.unbind())
+                    }
+                    ListDatapointsResponse::IdsOnly(_) => Err(convert_error(
+                        py,
+                        TensorZeroError::Other {
+                            source: TensorZeroCoreError::new(ErrorDetails::InvalidRequest {
+                                message: "`response_format: id` is not yet supported in the Python client"
+                                    .to_string(),
+                            })
```
virajmehta: same here

virajmehta (changes requested):
generally this LGTM but let's add the Python interface

PR #5819, opened 2026-01-23
Add xAI Responses API support 

### LINKED ISSUE #5774: Update xAI API integration to use Responses
The xAI API has deprecated their [chat completions](https://docs.x.ai/docs/guides/chat-completions) endpoint, which we use. We should move our integration over to the [Responses](https://docs.x.ai/docs/guides/chat) endpoint instead. 

We already do a similar parallel integration with OpenAI. We should similarly use this Responses API in a stateless way.

pratikbuilds:
Adds support for xAI's new Responses API while maintaining backward compatibility with the deprecated Chat Completions endpoint.

## Changes
- Add `XAIAPIType` enum with `chat_completions` (default) and `responses` variants
- Update `XAIProvider` to support both API types in streaming and non-streaming modes
- Reuse OpenAI Responses implementation for xAI Responses API
- Change visibility of OpenAI Responses types to `pub(crate)` for cross-provider reuse
- Add test configuration for Responses API (`grok_4_1_fast_non_reasoning_responses`)

## Testing
- ✅ `cargo check` passes
- ✅ `cargo fmt` passes
- ✅ `cargo clippy --all-targets --all-features -- -D warnings` passes
- Test configuration added for the new API type

Fixes #5774


GabrielBianconi:
Hi @pratikbuilds - unfortunately the xAI implementation has evolved so this won't be able to merge. I'll close it for now. Thank you for the submission.

PR #5939, opened 2026-01-28
Make postgres writes non-blocking

shuyangli:
We should not wait for Postgres writes if we're in a dual-write mode.

Step towards #5691.

### DIFF
--- a/tensorzero-core/src/db/delegating_connection.rs
+++ b/tensorzero-core/src/db/delegating_connection.rs
@@ -5,6 +5,7 @@
 //! feature flags.
 
 use async_trait::async_trait;
+use tokio_util::task::TaskTracker;
 use uuid::Uuid;
 
 use crate::db::TimeWindow;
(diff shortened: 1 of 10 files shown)


tensorzero-core/src/db/delegating_connection.rs
```diff
@@ -167,68 +174,93 @@ impl FeedbackQueries for DelegatingDatabaseConnection {
         &self,
         row: &BooleanMetricFeedbackInsert,
     ) -> Result<(), Error> {
-        self.clickhouse.insert_boolean_feedback(row).await?;
+        let clickhouse_future = self.clickhouse.insert_boolean_feedback(row);
```
Aaron1011: Is it intentional that the ClickHouse write is still blocking?

shuyangli:
Unfortunately I will keep this open for a bit - tests will be more annoying if we need to start waiting for postgres writes. Will revisit later

PR #6067, opened 2026-02-02
Wire up postgres e2e tests for dataset endpoints

shuyangli:
Step towards #5691.

> [!NOTE]
> **Medium Risk**
> Touches dataset endpoint wiring and Postgres datapoint serialization/parsing (notably `staled_at` datetime and tag null/empty behavior), which could change API-visible results and data written to Postgres alongside ClickHouse.
> 
> **Overview**
> **Routes dataset endpoint DB access through `DelegatingDatabaseConnection`** so `/v1/datasets/.../from_inferences` and `/v1/datasets/.../datapoints/metadata` operate against the ClickHouse+Postgres delegating layer rather than ClickHouse-only.
> 
> **Improves Postgres dataset compatibility** by parsing ClickHouse-formatted datetime strings for `staled_at` using an explicit `%Y-%m-%d %H:%M:%S%.6f` parser (with a unit test), and by always returning `tags` as `Some(map)` even when empty.
> 
> **Enables Postgres in dataset E2E coverage** by adding `DelegatingDatabaseConnection::new_for_e2e_test()` plus `TestDatabaseHelpers` support, then updating dataset endpoint E2E tests to use the delegating DB (removing ClickHouse-only setup, skip guards, and a couple of ClickHouse-specific assertions).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6ae41e215093c8d1cf4c55b131c2dbd6144dc757. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tensorzero-core/src/db/delegating_connection.rs
+++ b/tensorzero-core/src/db/delegating_connection.rs
@@ -509,7 +509,6 @@ impl DatasetQueries for DelegatingDatabaseConnection {
         Ok(results)
     }
 }
-
 #[async_trait]
 impl BatchInferenceQueries for DelegatingDatabaseConnection {
     // ===== Read methods: delegate based on ENABLE_POSTGRES_READ =====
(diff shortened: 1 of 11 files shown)


tensorzero-core/src/db/postgres/dataset_queries.rs
```diff
@@ -35,6 +35,22 @@ use crate::tool::wire::ToolChoice;
 
 use super::PostgresConnectionInfo;
 
+// ClickHouse datetime format used in the codebase: "%Y-%m-%d %H:%M:%S%.6f"
+// This format doesn't have timezone info, so we parse it as NaiveDateTime and assume UTC.
+const CLICKHOUSE_DATETIME_FORMAT: &str = "%Y-%m-%d %H:%M:%S%.6f";
```
shuyangli: This is because we currently use the Clickhouse data string in the StoredChat/JsonInferenceDatapoint types. We should rip it out and represent it as a native date type that gets serialized before writing/reading.

tensorzero-core/src/db/delegating_connection.rs
```diff
@@ -629,3 +628,40 @@ impl ModelInferenceQueries for DelegatingDatabaseConnection {
...
+            let clickhouse = get_clickhouse().await;
+            let postgres = get_postgres().await;
+            Self::new(clickhouse, postgres)
+        }
+    }
+
+    #[async_trait]
+    impl TestDatabaseHelpers for DelegatingDatabaseConnection {
+        async fn flush_pending_writes(&self) {
+            if ENABLE_POSTGRES_READ.get() || ENABLE_POSTGRES_WRITE.get() {
```
Aaron1011: Shouldn't we also flush clickhouse writes if only postgres reads are enabled?

PR #6186, opened 2026-02-05
Extract VariantsSection with independent Suspense boundary

simeonlee:
## Summary
- Extracts the variants table into its own `VariantsSection` component with independent Suspense boundary
- Variants now stream independently — they render as soon as their data loads, isolated from other sections
- Includes dedicated skeleton (3-row table) and error state (table with error notice)
- Remaining sections still share a single combined Suspense boundary

Part 3 of the function detail page streaming refactor (split from #6082). Stacks on #6181.


simeonlee:
Superseded

PR #6190, opened 2026-02-05
Extract InferencesSection with independent Suspense boundary

simeonlee:
## Summary
- Extracts the inferences table into its own `InferencesSection` component with independent Suspense boundary
- Includes dedicated skeleton (5-row table + disabled pagination) and error state (table with error notice)
- Removes the combined `fetchAllFunctionDetailData` wrapper and monolithic `SectionsContent`/`SectionsSkeleton`/`SectionsErrorState` components
- All 5 sections now stream independently with isolated error boundaries

This completes the function detail page streaming refactor (split from #6082). Stacks on #6189.


simeonlee:
Superseded

PR #6210, opened 2026-02-06
Implement config and deployment queries in Postgres

shuyangli:
A step towards #5691


tensorzero-core/src/db/clickhouse/mod.rs
```diff
@@ -372,6 +372,68 @@ impl ConfigQueries for ClickHouseConnectionInfo {
 
         ConfigSnapshot::from_stored(&row.config, row.extra_templates, row.tags, &snapshot_hash)
     }
+
+    async fn write_config_snapshot(&self, snapshot: &ConfigSnapshot) -> Result<(), Error> {
```
shuyangli: can you add unit tests for these? look at other db/clickhouse files for examples. use mocks.

tensorzero-core/src/db/postgres/config_queries.rs
```diff
@@ -0,0 +1,117 @@
...
+        let row = sqlx::query(
+            r"SELECT config, extra_templates, tags
+               FROM tensorzero.config_snapshots
+               WHERE hash = $1
+               LIMIT 1",
+        )
+        .bind(snapshot_hash.as_bytes())
+        .fetch_optional(pool)
+        .await
+        .map_err(|e| {
```
shuyangli: no need to map_err - you can just return ? and let the exisitng impl From\<sqlx::Error> do the job.

gateway/src/main.rs
```diff
@@ -71,6 +74,34 @@ async fn handle_disable_api_key(public_id: &str) -> Result<(), Box<dyn std::erro
...
+        return;
+    }
+
+    if clickhouse.client_type() != ClickHouseClientType::Production {
+        return;
+    }
+    if !matches!(postgres, PostgresConnectionInfo::Enabled { .. }) {
+        return;
+    }
+    task_tracker.spawn(async move {
```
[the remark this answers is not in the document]
shuyangli: now we're doing this synchronously during setup_howdy just before deployment ID is used.

Aaron1011 (approved):

PR #6214, opened 2026-02-07
Feat helm support config as secret

rguilmont:
Resolve #6277

I had this use case where i needed, in our production system, to mount the configuration from an existing secrets. Even though that might not be very common, it can be useful !

## Summary
 
- Adds Helm support for extra Kubernetes objects via extraObjects (handy for ExternalSecret / SealedSecret).
- Adds optional gateway config-from-Secret mount (`configFromExternalSecret`) 
- Adds `gateway.annotations` and documents the new values in the Helm README + gateway deployment docs.

### DIFF
--- a/docs/deployment/tensorzero-gateway.mdx
+++ b/docs/deployment/tensorzero-gateway.mdx
@@ -72,6 +72,8 @@ You can use it to run TensorZero in Kubernetes.
 
 The chart is available on [ArtifactHub](https://artifacthub.io/packages/helm/tensorzero/tensorzero).
 
+The chart supports mounting `tensorzero.toml` from a `Secret` (see `configFromExternalSecret`) and rendering additional objects like `ExternalSecret` / `SealedSecret` resources (see `extraObjects`).
+
 </Accordion>
 
 <Accordion title="Build from source">
--- a/examples/production-deployment-k8s-helm/README.md
+++ b/examples/production-deployment-k8s-helm/README.md
@@ -65,6 +65,7 @@ The following table lists the configurable parameters of the chart and their def
 
 | Parameter                    | Description                      | Default                                                                |
 | ---------------------------- | -------------------------------- | ---------------------------------------------------------------------- |
+| `gateway.annotations`        | Annotations for gateway Deployment | `{}`                                                                 |
 | `gateway.replicaCount`       | Number of gateway replicas       | `1`                                                                    |
 | `gateway.serviceAccountName` | Service account for gateway pods | `""`                                                                   |
 | `gateway.image.repository`   | Gateway image repository         | `tensorzero/gateway`                                                   |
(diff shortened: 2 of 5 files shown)


GabrielBianconi:
Pasting Claude's review below, for future reference:

```
PR #6214: Helm Support for Config as Secret

  Overview

  This PR adds three features to the TensorZero Helm chart:
  1. configFromExternalSecret — mount tensorzero.toml from a Kubernetes Secret instead of (or alongside) the ConfigMap
  2. extraObjects — render arbitrary additional Kubernetes objects (e.g., ExternalSecret, SealedSecret)
  3. gateway.annotations — add annotations to the gateway Deployment

  Issues & Suggestions

  1. Security: extraObjects uses tpl — arbitrary template execution

  extra-objects.yaml:3:
  {{ tpl (toYaml .) $ }}

  Using tpl on user-provided values allows arbitrary Helm template execution. This is a common pattern in charts
  (e.g., Bitnami), but it should be documented as intentional since it means any Helm function (include, lookup, etc.)
   can be called from values. This is fine if that's the intended use case, but worth being explicit about.

  2. configFromExternalSecret doesn't actually replace the ConfigMap — it adds a second mount

  When configFromExternalSecret.enabled is true, the ConfigMap volume is still mounted at /app/config. The secret
  volume is mounted at a separate path (default /app/config-secret/). The user then has to change gateway.args to
  point to the secret mount path.

  This is workable but has implications:
  - The ConfigMap is always created and mounted, even when unused. It would be cleaner to conditionally skip the
  ConfigMap volume mount when configFromExternalSecret.enabled is true.
  - The naming is misleading — configFromExternalSecret implies it replaces the ConfigMap-based config, but it
  actually adds an additional secret volume alongside the ConfigMap. Consider renaming to something like configSecret
  or additionalConfigSecret.

  3. configFromExternalSecret.secretName default collides with gateway.additionalEnv.secretName

  Both default to "tensorzero-secret" (values.yaml:39 and values.yaml:179). If a user enables configFromExternalSecret
   without changing the secret name, the chart will mount the environment variable secret as the config volume. These
  should have distinct default names (e.g., tensorzero-config-secret for the config one).

  4. Unrelated formatting changes

  The diff includes two unrelated cosmetic changes:
  - command: ["gateway"] → command: [ "gateway" ] (whitespace inside brackets)
  - Removed comment # Optional: Specify non-root user and group IDs from the UI securityContext

  These should be in a separate commit or removed from this PR.

  5. gateway.annotations is on the Deployment, not the Pod

  The annotations are placed on the Deployment metadata (gateway.yaml:8-11), not on the Pod template. This is fine for
   some use cases (e.g., Flux/ArgoCD notifications), but most annotation use cases in Kubernetes (Prometheus scraping,
   Istio sidecar injection, Vault agent injection) need pod-level annotations. Consider whether this should be
  gateway.podAnnotations instead (or in addition). Many popular charts distinguish between annotations (on the
  Deployment) and podAnnotations (on the Pod template).

  6. The README table doesn't document new top-level values

  The table in README.md documents gateway.annotations but doesn't include configFromExternalSecret.* or extraObjects
  in the parameter table. The separate prose sections are good, but the parameter table should be complete for
  discoverability.

  7. Schema for configFromExternalSecret lacks required fields

  In values.schema.json, configFromExternalSecret defines enabled, mountPath, and secretName but doesn't mark any as
  required. If enabled is true, both secretName and mountPath should be required — otherwise the chart will silently
  render broken YAML (empty secretName or mountPath).

  8. Config checksum annotation won't trigger rollout on secret changes

  gateway.yaml:21:
  checksum/config: {{ toYaml .Values.configMap | sha256sum }}

  This only checksums the ConfigMap values. When using configFromExternalSecret, changes to the external secret won't
  trigger a pod rollout. Consider adding a second checksum annotation when the secret config is enabled, or
  documenting this limitation.

  Summary

  The feature itself is useful and well-scoped. The main concerns are:

  - Name collision on default secretName between env vars and config — this will cause user confusion
  - ConfigMap still mounted when unused — minor waste / potential confusion
  - Annotations on Deployment vs Pod — may not match common use cases
  - Unrelated formatting changes mixed in
```

PR #6257, opened 2026-02-10
Add trigram indices for input and output

shuyangli:
A step towards #5691.

We've decided that for simplicity we will offer a trigram index over the request/response bodies in Postgres for full-text search for now. This will work for small-medium scale deployments with reasonable TTLs.

This is not a migration because we don't want to fail for people who only use Postgres for auth and not observability. We also build the indices concurrently for partitioned tables so people can add observability later.

---

> [!NOTE]
> **Medium Risk**
> Touches Postgres startup/migration flow and runs `CREATE EXTENSION`/`CREATE INDEX CONCURRENTLY`/`ATTACH PARTITION` statements, which can have operational/performance implications on large databases. Logic is idempotent and largely additive, but failures/misconfiguration now produce new warnings and checks.
> 
> **Overview**
> Adds a new `postgres_setup` module that centralizes Postgres initialization: keeps the existing idempotent `pg_cron` setup/checks and adds **optional `pg_trgm` setup** plus creation/attachment of trigram GIN indexes for `input`/`output` across inference and datapoint tables (handling partitioned tables by indexing partitions concurrently and attaching them to a parent index).
> 
> `manual_run_postgres_migrations_with_url` now runs trigram index setup on every migration run and emits consolidated warnings when either `pg_cron` or trigram indexing is misconfigured, and e2e tests are updated/expanded to cover trigram index existence, idempotency, and partition attachment; the old standalone `pgcron` module/tests are removed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dbea1f2b650a1f056668717574a4529674663775. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tensorzero-core/src/db/postgres/postgres_setup.rs
+++ b/tensorzero-core/src/db/postgres/postgres_setup.rs
@@ -211,28 +211,37 @@ async fn get_partitions(pool: &PgPool, table: &str) -> Result<Vec<String>, Error
     })
 }
 
-/// Returns whether a given index (by name in `tensorzero` schema) is already
-/// attached as a partition of some parent index via `pg_inherits`.
-async fn is_index_attached(pool: &PgPool, index_name: &str) -> Result<bool, Error> {
-    sqlx::query_scalar(
+/// Returns the set of partition table names that already have an index attached to
+/// the given parent index. This handles both our explicitly named indexes and
+/// Postgres auto-created ones, and performs a single database round-trip.
+async fn get_partitions_with_attached_index(
+    pool: &PgPool,
+    parent_index_name: &str,
+) -> Result<std::collections::HashSet<String>, Error> {
+    let rows: Vec<String> = sqlx::query_scalar(
         r"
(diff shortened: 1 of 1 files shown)


Aaron1011 (approved):

tensorzero-core/src/db/postgres/postgres_setup.rs
```diff
@@ -0,0 +1,366 @@
...
+    partitions: &[String],
+) -> Result<(), Error> {
+    let parent_index = format!("idx_{table}_{column}_trgm");
+    for partition in partitions {
+        let partition_index = format!("idx_{partition}_{column}_trgm");
+        if is_index_attached(pool, &partition_index).await? {
+            continue;
+        }
+        let sql = format!(
+            "ALTER INDEX tensorzero.{parent_index} \
```
Aaron1011: Can you double-check that this is idempotent?

PR #6286, opened 2026-02-11
Add cost tracking for model providers

AntoineToussaint:
## Summary

Adds cost tracking infrastructure for model and embedding providers. Cost is computed at inference time from raw provider responses using user-configured JSON Pointer mappings, then stored in both ClickHouse and Postgres.

### New module: `cost.rs`

- `Cost` type alias for `rust_decimal::Decimal` (stored as `Decimal(18, 9)` / `NUMERIC(18, 9)`)
- **Uninitialized → Normalized config pattern**:
  - `UninitializedCostConfigEntry` / `UninitializedCostRate` — user-facing types with `cost_per_million` and `cost_per_unit`, deserialized from TOML/JSON
  - `CostConfigEntry` / `CostRate` — normalized runtime types (always per-unit), converted at config load time via `TryFrom` / `From` impls
- `CostPointerConfig` (unified or split streaming/non-streaming pointers) — stored as `String` in user-facing config, validated and parsed into `json_pointer::JsonPointer` at config load time
- `ResponseMode` enum (`NonStreaming` / `Streaming`) instead of boolean flag
- `compute_cost_from_response()` / `compute_cost_from_json()` — extracts values via JSON Pointers, applies rates, sums contributions
- `CostConfigEntry::compute_contribution()` — per-entry cost computation, easily extensible for future expression-based cost models
- Custom `decimal_from_value` serde visitor — formats f64 to shortest decimal string (Ryu algorithm) then parses to `Decimal`, avoiding binary float precision loss (e.g. `0.1` in TOML → exactly `Decimal(0.1)`)
- Negative rates are allowed (e.g., caching discounts); negative **totals** log a warning and return `None`

### Configuration

- New optional `cost: Option<CostConfig>` field on model providers and embedding providers
- Pointers validated at config load time using `json_pointer` crate
- Example:
  ```toml
  [models."gpt-4o-mini-2024-07-18".providers.openai]
  type = "openai"
  model_name = "gpt-4o-mini-2024-07-18"
  cost = [
    { pointer = "/usage/prompt_tokens", cost_per_million = 0.15, required = true },
    { pointer = "/usage/completion_tokens", cost_per_million = 0.60, required = true },
  ]
  ```

### Database

- **ClickHouse migration 0047**: `ALTER TABLE ModelInference ADD COLUMN cost Nullable(Decimal(18, 9))`
- **Postgres migration**: `ALTER TABLE tensorzero.model_inferences ADD COLUMN cost NUMERIC(18, 9)`
- All SELECT/INSERT queries for `model_inferences` updated to include `cost`

### Inference pipeline

- **Non-streaming**: cost computed immediately after provider response from `raw_response`
- **Streaming**: cost computed per-chunk inside stream processing loop, enriching chunks before yielding to downstream consumers
- **Embeddings**: cost computed after embedding response (always non-streaming)
- `Usage` struct gains `cost: Option<Cost>` field (derives `Copy`)
- **Streaming aggregation**: uses `streaming_max` for cost (same strategy as tokens)
- **Cross-model aggregation**: uses `sum_or_poison` — sums cost when all inferences have cost; any `None` → total `None` (poison semantics)

### API / wire types

- `ModelInference` (internal API): new `cost: Option<Cost>` (omitted when `None`)
- `StoredModelInference`: new `cost: Option<Cost>`
- `OpenAICompatibleUsage`: new `tensorzero_cost: Option<Decimal>` extension field

### Provider changes

All 16 providers updated to initialize `cost: None` in `Usage` construction (cost is computed later from `raw_response`): Anthropic, AWS Bedrock, Dummy, Fireworks, GCP Vertex Anthropic, GCP Vertex Gemini, Google AI Studio Gemini, Groq, Mistral, OpenAI, OpenAI Responses, OpenRouter, TGI, Together, XAI, relay.

### Variant changes

`cost` field threaded through `best_of_n_sampling`, `chat_completion`, `mixture_of_n`, and variant `mod.rs`.

### Optimizer changes

All 4 optimizers updated with `cost: None`: DICL, Fireworks SFT, OpenAI, Together SFT.

### Dependencies

- `rust_decimal = { version = "1.37", features = ["serde-float"] }` (new)
- `json_pointer` (new, for pointer validation)
- `sqlx` feature `"rust_decimal"` added

### Tests

**69 cost-related tests passing** (unit + E2E):

#### Unit tests in `cost.rs` (45 tests)

- **Rate conversion**: per-million → per-unit normalization, per-unit passthrough
- **Cost computation**: basic token pricing, per-unit pricing (web searches), required/optional missing fields, negative totals, empty config, split pointers (streaming vs non-streaming), unparseable values, string-encoded numbers, booleans, zero tokens, float values, mixed positive/negative (caching discount), discount exceeding base cost
- **Response parsing**: valid JSON, invalid JSON
- **TOML deserialization**: unified pointers, per-unit rates, split pointers, invalid pointers, invalid split pointers, missing rates, missing pointers, invalid cost values, exact decimal precision (0.1, 0.3), string-quoted decimals, negative costs, both rates present, empty pointer string
- **Config normalization**: per-million → per-unit, per-unit passthrough, `load_cost_config` with valid config
- **JSON Pointer validation**: valid, invalid (missing leading slash)
- **Full pipeline** (TOML → normalize → provider JSON → cost): OpenAI-style, Anthropic-style with cache, required field missing, per-unit web search, invalid JSON, zero tokens, large token counts (GPT-4), Gemini-style nested usage, direct cost passthrough
- **Warning assertion**: `test_compute_cost_required_missing_logs_warning` uses `capture_logs()` to verify warning is logged when required pointer is missing

#### Unit tests in `usage.rs` (20 tests)

- **Streaming aggregation**: empty, single chunk, cumulative (Anthropic-style), final-only (OpenAI-style), all None, mixed None/Some, non-cumulative detection (debug_assert), Anthropic real-world pattern
- **Cross-model aggregation**: empty, single, multiple all-Some, None propagation (poison semantics), both None propagate, all None
- **Cost aggregation**: both present (sum), one missing poisons total, missing first poisons total, all None, zero is present (not missing), empty returns zero, three-way mix poisons (mixture-of-N scenario)

#### E2E tests (16 tests)

- **Simple cost** (non-streaming + streaming): cost present in API response, cost stored correctly in ClickHouse
- **DICL poison semantics** (non-streaming + streaming): model has cost config, embedding model doesn't → total cost is `None`
- **Best-of-N poison semantics** (non-streaming + streaming): candidates have cost, evaluator doesn't → total cost is `None`
- **Best-of-N all-cost** (non-streaming + streaming): all providers (candidates + evaluator) have cost config → total cost is sum
- **Mixture-of-N poison semantics** (non-streaming + streaming): one candidate model lacks cost → total cost is `None`
- **Mixture-of-N all-cost** (non-streaming + streaming): all providers (candidates + fuser) have cost config → total cost is sum

#### Test infrastructure additions

- `best_of_n_0_with_usage` dummy provider: returns evaluator content with standard usage fields for cost tracking tests
- `dummy_dicl` variant: uses `dummy-embedding-model` (no cost config) for DICL poison semantics tests
- `cost_test_best_of_n`, `cost_test_best_of_n_all_cost`, `cost_test_mixture_of_n`, `cost_test_mixture_of_n_all_cost` functions in E2E config

## Test plan

- [x] `cargo check --all-targets --all-features`
- [x] `cargo fmt --check`
- [x] Unit tests: 65 tests passing (`cost.rs` + `usage.rs`)
- [x] E2E tests: 69 tests run, 69 passed (including 16 cost-specific E2E tests)
- [x] All existing tests continue to pass (6387 skipped = non-cost tests filtered out)


tensorzero-core/src/db/clickhouse/migration_manager/migrations/migration_0046.rs
```diff
@@ -1,80 +1,46 @@
...
-            let _ = self
-                .clickhouse
-                .run_query_synchronous_no_params(materialize_index_query)
-                .await?;
-        }
+        self.clickhouse
+            .run_query_synchronous_no_params(format!(
+                "ALTER TABLE ModelInference{on_cluster_name} ADD COLUMN IF NOT EXISTS cost Nullable(Decimal(18, 9))"
+            ))
+            .await?;
```
virajmehta: we should also aggregate the cost for all ModelInferences which happen in a ChatInference or JsonInference so we'll need a new column on those 2 tables as well

tensorzero-core/src/endpoints/openai_compatible/types/usage.rs
```diff
@@ -15,6 +16,9 @@ pub struct OpenAICompatibleUsage {
     pub completion_tokens: Option<u32>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub total_tokens: Option<u32>,
+    /// Cost of this inference in dollars (TensorZero extension).
```
virajmehta: probably also don't need the (t0 extension)

tensorzero-core/src/cost.rs
```diff
@@ -0,0 +1,838 @@
...
+                } else {
+                    pointer_nonstreaming
+                }
+            }
+        }
+    }
+}
+
+/// The rate to apply to an extracted value.
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
```
virajmehta: I don't think we need to derive Deserialize here

virajmehta: just kidding, we may need this only because it lives in UninitializedConfig (which we have a separate machine-friendly storage format for in the database for snapshotting)

tensorzero-core/src/embeddings.rs
```diff
@@ -626,6 +632,14 @@ impl EmbeddingProviderInfo {
...
             response_fut.await?
         };
+        // Compute cost from the raw response if cost config is provided
+        if let Some(cost_config) = &self.cost {
+            response.usage.cost = crate::cost::compute_cost_from_response(
+                &response.raw_response,
+                cost_config,
+                false, // embeddings are never streaming
+            );
+        }
```
GabrielBianconi: Might be less error prone to compute this closer to the source rather than setting it to None then modifying later

GabrielBianconi (changes requested):
I'd like to see tests (possibly E2E tests) that cover:

- Gateway relay
- Advanced variant types (best/mixture of N, DICL)

^ non-streaming and streaming versions of all of the above

tensorzero-core/src/cost.rs
```diff
@@ -0,0 +1,838 @@
...
+    if total_cost < Decimal::ZERO {
+        tracing::warn!("Computed cost is negative ({total_cost}). Cost will be unavailable.");
+        return None;
+    }
+
+    Some(total_cost)
+}
+
+/// Convert a JSON value to a Decimal, handling integer, float, and string representations.
+fn json_value_to_decimal(value: &serde_json::Value) -> Option<Decimal> {
```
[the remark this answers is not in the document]
virajmehta: one solution is to ask users to quote the number 🤮

GabrielBianconi (changes requested):
Following up on my comment above, in addition to success cases, we'll want to test full error cases (missing required cost info) data and mixed cases (e.g. DICL - embeddings doesn't have proper cost - None; model inference does; combined should downgrade to missing).

tensorzero-core/src/cost.rs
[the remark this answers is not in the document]
GabrielBianconi: https://docs.rs/rust_decimal/latest/rust_decimal/serde/index.html

tensorzero-core/src/cost.rs
```diff
@@ -0,0 +1,838 @@
...
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[cfg_attr(feature = "ts-bindings", derive(ts_rs::TS))]
+#[serde(untagged)]
+pub enum CostPointerConfig {
+    /// A single pointer used for both streaming and non-streaming responses.
+    Unified { pointer: String },
+    /// Separate pointers for streaming and non-streaming responses.
+    Split {
+        pointer_nonstreaming: String,
+        pointer_streaming: String,
```
virajmehta: should we hold these as a json_pointer instead of string? will simplify validation

tensorzero-core/src/cost.rs
```diff
@@ -0,0 +1,838 @@
...
+                None
+            }
+        }
+        serde_json::Value::String(s) => s.parse::<Decimal>().ok(),
+        _ => None,
+    }
+}
+
+/// Validate a cost configuration, returning an error if it's invalid.
+pub fn validate_cost_config(cost_config: &[CostConfigEntry]) -> Result<(), Error> {
```
virajmehta: if we hold the config pointers as `json_pointer::JsonPointer` then we should not have to manually validate

tensorzero-core/src/cost.rs
```diff
@@ -0,0 +1,838 @@
...
+    PerMillion {
+        #[cfg_attr(feature = "ts-bindings", ts(type = "number"))]
+        cost_per_million: Decimal,
+    },
+    /// Cost per single unit (e.g., cost per web search).
+    PerUnit {
+        #[cfg_attr(feature = "ts-bindings", ts(type = "number"))]
+        cost_per_unit: Decimal,
+    },
+}
```
virajmehta: we should have an UninitializedCostConfig/Rate that looks like this and then as the config is converted from UnitializedConfig -> Config we should `impl From UninitializedCostConfig/Rate for CostConfig/Rate` and put the conversions of each of these enum arms into a single normalized representation (most likely cost per unit). That way, the runtime logic is clean but the deserialization format is user-friendly. This also avoids having to write a custom deserializer (hard to round-trip and maintain).

tensorzero-core/src/model.rs
```diff
@@ -871,11 +883,17 @@ pub struct ModelProvider {
...
+    /// Cost configuration for computing inference cost from raw provider responses.
+    #[cfg_attr(feature = "ts-bindings", ts(skip))]
+    pub cost: Option<crate::cost::CostConfig>,
 }
 
 impl ModelProvider {
     fn validate(&self, global_outbound_http_timeout: &chrono::Duration) -> Result<(), Error> {
         self.timeouts.validate(global_outbound_http_timeout)?;
+        if let Some(cost_config) = &self.cost {
+            crate::cost::validate_cost_config(cost_config)?;
```
virajmehta: general style: I don't like the inline imports

tensorzero-core/src/db/clickhouse/migration_manager/migrations/migration_0046.rs
virajmehta: just kidding, after discussion we've decided to just store it here. This will reduce scope :)

AntoineToussaint:
recheck

tensorzero-core/src/embeddings.rs
AntoineToussaint: Update: refactored to a one-liner: `response.usage.cost = self.cost.as_ref().and_then(|cfg| compute_cost_from_response(&response.raw_response, cfg, false));` — cost is computed immediately after receiving the response.

tensorzero-core/src/cost.rs
AntoineToussaint: Addressed — added the `json-pointer` crate (`json_pointer::JsonPointer`) for RFC 6901 validation. The uninitialized config stores `String` (serde-friendly), while the normalized runtime type (`NormalizedCostPointerConfig`) stores a parsed `json_pointer::JsonPointer`. Validation happens at config load time via `parse_json_pointer()`.

tensorzero-core/src/cost.rs
virajmehta: we discussed this live but I think this should be the native `JsonPointer` type

tensorzero-core/src/cost.rs
```diff
@@ -0,0 +1,1496 @@
...
+            }
+
+            fn visit_str<E: de::Error>(self, v: &str) -> Result<Self::Value, E> {
+                v.parse::<Decimal>().map_err(E::custom)
+            }
+        }
+
+        deserializer.deserialize_any(DecimalVisitor)
+    }
+}
```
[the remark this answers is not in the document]
AntoineToussaint: Definitely on the ugly side but tests validate. Feels the "interface is correct", in the sense we use cost_..=0.1 and not string. The implmentation itself is on the hackish side (but tested) so maybe a follow up PR can go into the proper TOML parsing?

tensorzero-core/src/cost.rs
AntoineToussaint: Addressed — pointers are stored as `String` in the user-facing `UninitializedCostConfigEntry` (for easy Deserialize from TOML), then parsed and validated into `json_pointer::JsonPointer` at config load time via `parse_json_pointer()`. The runtime `NormalizedCostPointerConfig` stores validated `JsonPtr` values — no manual validation needed at runtime.

tensorzero-core/src/inference/types/usage.rs
```diff
@@ -101,54 +147,33 @@ where
...
+                chunk_usage.input_tokens,
+                Some("input_tokens"),
+            ),
+            output_tokens: streaming_max(
+                acc.output_tokens,
+                chunk_usage.output_tokens,
+                Some("output_tokens"),
+            ),
+            // Cost uses the same max strategy but without the non-cumulative warning
+            cost: streaming_max(acc.cost, chunk_usage.cost, None),
```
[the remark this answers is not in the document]
virajmehta: I believe this is unavoidable given existing provider semantics for cost.

virajmehta (approved):

GabrielBianconi (dismissed):

GabrielBianconi (approved):

PR #6309, opened 2026-02-12
Make chat and json inferences support optional data types

### LINKED ISSUE #6297: Make input/output/etc optional in data model
For ZDR and storage reasons

shuyangli:
Fixes #6297.

This doesn't actually change how data is stored, but it allows us to drop the data (and only keep metadata) and keep downstream code working.

UI looks fine:


clients/python/tests/test_list_inferences.py
virajmehta: can we test the case (can be in Rust too) where it should be None?

tensorzero-core/src/db/clickhouse/query_builder/mod.rs
virajmehta: can we add a test for the None case here?

tensorzero-core/src/db/postgres/inference_queries.rs
```diff
@@ -1118,53 +1118,56 @@ impl<'r> sqlx::FromRow<'r, sqlx::postgres::PgRow> for StoredChatInferenceDatabas
         let variant_name: String = row.try_get("variant_name")?;
         let episode_id: Uuid = row.try_get("episode_id")?;
         let timestamp: DateTime<Utc> = row.try_get("timestamp")?;
-        let input: Json<StoredInput> = row.try_get("input")?;
-        let output: Json<Vec<ContentBlockChatOutput>> = row.try_get("output")?;
+        let input: Option<Json<StoredInput>> = row.try_get("input").unwrap_or(None);
```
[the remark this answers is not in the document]
virajmehta: @shuyangli I think this is correct

virajmehta (changes requested):
can you also look at how this renders in the UI and make sure it doesn't explode?

PR #6310, opened 2026-02-12
Incremental latency quantiles computation

shuyangli:
Instead of calculating and storing exact percentiles in a materialized view, this buckets request latencies into `floor(log2(latency) * 64` buckets (so total number of buckets is practically bounded) to give us reasonable estimates of p99.9 latency. We can use midpoint of the bucket as the estimated latency, and if the tail latency is 1 minute, our worst-case error from estimation is ~300ms.

At query time, quantiles are computed from histogram CDFs:

1. Aggregate bucket counts over the requested window per (model, metric, bucket_id).
2. Build cumulative counts ordered by bucket_id.
3. For each target quantile, compute rank target = 1 + quantile * (total_count - 1).
4. Pick the first bucket where cumulative count >= rank target.
5. Interpolate within that bucket in log-space between bucket bounds to estimate the quantile value.

Also we should consider not returning p0.1 and p99.9 when count is small, because the error will be high.

A step towards #5691.

---

> [!NOTE]
> **Medium Risk**
> Touches core Postgres schema and background refresh scheduling for dashboard latency metrics; mistakes could lead to stale/incorrect quantiles or increased DB load during refresh windows.
> 
> **Overview**
> Replaces the `model_latency_quantiles*` materialized views with **incrementally maintained sparse latency histograms** (minute + hour rollups) and **regular views** that compute approximate quantiles from those histograms.
> 
> Adds new Postgres tables/functions to bucket latencies (log2 buckets) and refresh rollups incrementally with persisted watermarks, and updates pg_cron setup/validation + e2e tests + fixture scripts to run the new `tensorzero_refresh_model_latency_histograms_incremental` job instead of `REFRESH MATERIALIZED VIEW`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ebf602739ce0c98aa5170d07de7cd809223c9dca. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ui/fixtures/load_fixtures_postgres.sh
+++ b/ui/fixtures/load_fixtures_postgres.sh
@@ -418,7 +418,10 @@ psql -q "$POSTGRES_URL" <<EOF
 SELECT tensorzero.refresh_model_provider_statistics_incremental(
     full_refresh => TRUE
 );
-SELECT tensorzero.refresh_model_latency_histograms_incremental(
+SELECT tensorzero.refresh_model_latency_histogram_minute_incremental(
+    full_refresh => TRUE
+);
+SELECT tensorzero.refresh_model_latency_histogram_hour_incremental(
     full_refresh => TRUE
 );
 EOF


shuyangli:
@bugbot review

tensorzero-core/src/db/postgres/migrations/20260212110000_incremental_model_latency_quantiles.sql
```diff
@@ -0,0 +1,730 @@
...
+)
+RETURNS INTEGER
+LANGUAGE sql
+IMMUTABLE
+AS $$
+SELECT
+    CASE
+        WHEN latency_ms IS NULL THEN NULL
+        WHEN latency_ms <= 0 THEN -1
+        ELSE FLOOR((LN(GREATEST(latency_ms, 1)::DOUBLE PRECISION) / LN(2.0)) * buckets_per_power_of_two)::INTEGER
```
virajmehta: if I'm understanding this correctly, the worst case error is if you get a bunch of data points with latency $$2^k - \epsilon$$ (the widest bucket for a particular power of 2). In that case, we'll run into ~1.1% error regardless of $$k$$. I think this is acceptable.  Please add a code comment to that effect.

tensorzero-core/src/db/postgres/migrations/20260212110000_incremental_model_latency_quantiles.sql
```diff
@@ -0,0 +1,730 @@
...
+                        PARTITION BY h.model_name, h.metric
+                        ORDER BY h.bucket_id
+                    ) - h.bucket_count AS cumulative_before,
+                    CASE
+                        WHEN h.bucket_id < 0 THEN 0.0
+                        ELSE POWER(2.0, h.bucket_id::DOUBLE PRECISION / p.buckets_per_power_of_two)
+                    END AS lower_ms,
+                    CASE
+                        WHEN h.bucket_id < 0 THEN 1.0
+                        ELSE POWER(2.0, (h.bucket_id + 1)::DOUBLE PRECISION / p.buckets_per_power_of_two)
```
virajmehta: it is really unfortunate that we can't unit test logic like this to make sure its doing what we think

shuyangli:
shifted the bucketing and quantiles logic into rust, now the migration is only for rolling up raw data into minutes/hours tables

Aaron1011 (approved):

PR #6320, opened 2026-02-13
Allow unknown fields for config over the wire

shuyangli:
Autopilot in staging could consume a local gateway whose configs may have new fields (unknown to autopilot). We need to make sure over-the-wire serde would not fail on these unknown fields.

We still fail config loading from toml with serde_ignored - if there's any ignored field, gateway startup should fail.

This is only a mitigation for the devex problem - for production purposes, autopilot should always be ahead of released gateways. We should not make backwards-incompatible changes to config shape - autopilot should be able to deserialize an old config shape. (We can put guardrails to say "you cannot use autopilot if gateway version is X behind" so we can eventually deprecate old config shapes, but need a documented decision.)

To get #6229 in:
- Submit this
- Update autopilot to pin against this hash
- Profit

---

> [!NOTE]
> **Medium Risk**
> Touches core config parsing semantics: unknown fields are now tolerated in some deserialization paths while TOML loading becomes stricter, which could change compatibility and error behavior for existing integrations.
> 
> **Overview**
> Config deserialization is relaxed for *over-the-wire* uses by removing `#[serde(deny_unknown_fields)]` from several gateway/observability-related config structs/enums, allowing newer gateways to send configs that older consumers can still parse.
> 
> To preserve strictness for local TOML startup, `UninitializedConfig::try_from(toml::Table)` now composes `serde_path_to_error` with `serde_ignored` to **fail fast on unknown keys** (with improved paths) and adds a test ensuring an unknown `[gateway]` field is rejected. Dependency updates add `serde_ignored` to the workspace and `tensorzero-core`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1792a1cf3fcba4549500cc15cbb896e9711e0d10. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5671,6 +5671,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "serde_ignored"
+version = "0.1.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "115dffd5f3853e06e746965a20dcbae6ee747ae30b543d91b0e089668bb07798"
+dependencies = [
+ "serde",
+ "serde_core",
+]
+
+[[package]]
 name = "serde_json"
 version = "1.0.149"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


shuyangli:
@bugbot review

PR #6346, opened 2026-02-15
Ensure Valkey inference caching does not evict rate limiting keys

shuyangli:
## Summary

This PR introduces support to make sure inference caching does not evict rate limiting keys under memory pressure with 2 options:

- Allow users to specify a dedicated Valkey instance for model inference caching, separate from the primary instance used for rate limiting. This allows operators to configure different eviction policies for each workload, preventing cache eviction from disrupting rate limiting state.
- If they want to use a single instance, we also increased the minimum TTL for rate limiting keys to 48 hours. By default the TTL for model inference cache is 24 hours, so `volatile-ttl` policy will not evict rate limiting keys with this configuration. This requires documentation to make sure people configures things properly (to come).

## Key Changes

- **New environment variable**: `TENSORZERO_VALKEY_CACHE_URL` allows pointing caching at a dedicated Valkey instance. When not set, caching falls back to the primary `TENSORZERO_VALKEY_URL` instance.
- **New** **`setup_valkey_cache()`** **function**: Handles initialization of the cache-specific Valkey connection, with fallback logic to reuse the primary connection if no dedicated cache URL is provided.
- **Updated** **`AppStateData`** **struct**: Added `valkey_cache_connection_info` field to track the cache-specific connection separately from the primary rate limiting connection.
- **Cache manager initialization**: Modified to use `valkey_cache_connection_info` instead of the primary `valkey_connection_info`, enabling independent eviction policies.
- **Rate limiting key TTL increase**: Increased minimum TTL for rate limiting keys from 1 hour to 48 hours in the Lua script. This ensures that under `volatile-ttl` eviction policy, rate limiting keys are evicted last (after the default 24-hour cache TTL).
- **Health check endpoint**: Updated to report the status of both the primary Valkey instance and the cache instance separately.
- **Documentation**: Added comprehensive guidance on using dedicated cache instances, eviction policies, and best practices for protecting rate limiting state.

## Implementation Details

- The separation is transparent to existing deployments: if `TENSORZERO_VALKEY_CACHE_URL` is not set, the cache uses the same connection as rate limiting.
- All initialization paths (embedded gateway, client builder, snapshot loading) have been updated to handle the new cache connection parameter.
- The change maintains backward compatibility while providing operators with fine-grained control over memory management and eviction behavior.

https://claude.ai/code/session_01EaNNAjFLoXj2iUgKDh6X2o

skip-if-edited-check: types are not changed.

### DIFF
--- a/docs/deployment/valkey-redis.mdx
+++ b/docs/deployment/valkey-redis.mdx
@@ -81,6 +81,21 @@ If both `TENSORZERO_VALKEY_URL` and `TENSORZERO_POSTGRES_URL` are set, the gatew
 
 </Warning>
 
+## Best Practices
+
+### Eviction Policy
+
+Rate limiting keys and cache entries have different eviction tolerance:
+
+- **Rate limiting** is a correctness concern — silent eviction resets rate limit state and allows clients to exceed their limits.
+- **Caching** is a performance concern — eviction simply causes a cache miss and the next request re-fetches from the provider.
+
+If you use a single Valkey instance for both, we recommend the `volatile-ttl` eviction policy.
+TensorZero sets rate limiting key TTLs to a minimum of 48 hours (compared to the default 24-hour cache TTL), so rate limiting keys are evicted last under memory pressure.
+Please configure cache TTL to be less than 48 hours to ensure correct eviction ordering.
+
+For stronger guarantees, use a [dedicated Valkey instance for caching](#dedicated-valkey-instance-for-caching-optional) with its own eviction policy, while keeping the primary instance configured with `noeviction`.
+
 ### Dedicated Valkey Instance for Caching (Optional)
 
 By default, TensorZero uses a single Valkey instance for both **rate limiting** and **model inference caching**.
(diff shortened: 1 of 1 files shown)


gateway/src/main.rs
```diff
@@ -379,6 +379,9 @@ async fn run() -> Result<(), ExitCode> {
     let valkey_enabled_pretty =
         get_valkey_status_string(&gateway_handle.app_state.valkey_connection_info);
     tracing::info!("├ Valkey: {valkey_enabled_pretty}");
+    let valkey_cache_enabled_pretty =
```
shuyangli: let's only print this if there's a separate `TENSORZERO_VALKEY_CACHE_URL`

clients/python/src/lib.rs
```diff
@@ -700,6 +702,7 @@ impl TensorZeroGateway {
             clickhouse_url,
             postgres_config: postgres_url.map(PostgresConfig::Url),
             valkey_url,
+            valkey_cache_url: None,
```
shuyangli: let's not expose a separate option for valkey_cache_url for embedded gateways. we can add a TODO, but for now all embedded gateways should just use `valkey_url` for both purposes.

docs/deployment/valkey-redis.mdx
```diff
@@ -81,8 +81,42 @@ If both `TENSORZERO_VALKEY_URL` and `TENSORZERO_POSTGRES_URL` are set, the gatew
...
 
+### Eviction Policy
+
+Rate limiting keys and cache entries have different eviction tolerance:
+
+- **Rate limiting** is a correctness concern — silent eviction resets rate limit state and allows clients to exceed their limits.
+- **Caching** is a performance concern — eviction simply causes a cache miss and the next request re-fetches from the provider.
+
+If you use a single Valkey instance for both, we recommend the `volatile-ttl` eviction policy.
+TensorZero sets rate limiting key TTLs to a minimum of 48 hours (compared to the default 24-hour cache TTL), so rate limiting keys are evicted last under memory pressure.
```
shuyangli: add another line saying please configure cache ttl to be less than 48 hours for eviction correctness

PR #6564, opened 2026-02-24
[STACKED] Make GEPA durable-only with per-iteration checkpointing

anndvision:
depends on #6561 
enables #6521 

## Summary

Break the monolithic GEPA iteration into 5 checkpointable sub-steps, add `TensorZeroClient` trait methods for each, and introduce `GepaSideInfo` / standalone tool visitor infrastructure.

### Key changes

- **Per-step checkpointing**: Each GEPA iteration is broken into 5 sub-steps (sample, eval-parent, analyze, mutate, eval-and-update) with serializable params/results between them, so the durable framework can checkpoint after each LLM call batch without inflating `claim_timeout`.
- **`GepaSideInfo`**: Train/val examples moved from `GepaToolParams` (LLM params) to `GepaSideInfo` (side info). Examples are infrastructure data, not LLM-generated.
- **`visit_standalone_task_tool`**: New visitor method for tools that bypass `ClientTaskToolWrapper`. `ToolNameCollector` implements as no-op (standalone tools aren't visible to autopilot). `LocalToolVisitor` registers directly with executor.
- **Remove `skip_publish`**: Field removed from `AutopilotSideInfo`. Standalone tools bypass the wrapper entirely instead.
- **Two-executor architecture**: Separate `variant_executor` and `judge_executor` for inference vs analysis/mutation calls.
- **RNG seed offset**: `from_checkpoint` offsets seed by iteration number to advance RNG state across checkpoint boundaries. Minibatch sampling uses a decorrelated seed derivation.
- **TensorZeroClient GEPA methods**: 7 new trait methods (`gepa_setup`, `gepa_iter_sample`, `gepa_iter_eval_parent`, `gepa_iter_analyze`, `gepa_iter_mutate`, `gepa_iter_eval_and_update`, `gepa_cleanup`) with embedded implementations and client stubs.
- **Boxed enum variants**: Sub-step result enums use `Box<Continue>` and `Box<GepaSkipIteration>` to avoid large stack allocations.
- **Derived `JsonSchema`**: `GepaToolParams` uses `#[derive(JsonSchema)]` instead of a manual impl.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo fmt`
- [x] `cargo clippy` (all packages, all targets)
- [x] `cargo test-unit-fast` (2578 tests pass)
- [x] Pre-commit hooks pass

---

> [!NOTE]
> **High Risk**
> High risk because it changes the core `Optimizer`/`JobHandle` APIs and optimization endpoints to accept `SpawnClient`, and converts GEPA from synchronous execution to durable task orchestration/polling.
> 
> **Overview**
> GEPA optimization is converted from an in-process synchronous optimizer to a **durable task**: `GEPAConfig::launch()` now spawns `gepa_optimization` via `SpawnClient` and returns a `GEPAJobHandle { task_id }`, while polling reads durable task status/results and maps them into `OptimizationJobInfo`.
> 
> The GEPA pipeline is split into **checkpointable sub-steps** (setup + 5 iteration steps + cleanup) with new serializable types (`gepa/types.rs` and `ParetoCheckpoint`), and `analyze`/`mutate` are refactored to use an `EvaluationsInferenceExecutor` abstraction (plus helpers like `build_inference_executor`).
> 
> To support durable execution, optimization workflow/HTTP handlers and the shared `Optimizer`/`JobHandle` traits now thread an optional `SpawnClient` through `launch_optimization*` and `poll_optimization`, and `TensorZeroClient` gains GEPA sub-step methods (embedded implementations; non-embedded return `NotSupported`). Autopilot tooling adds `visit_standalone_task_tool` to register tools that bypass autopilot wrappers, and tests/bindings are updated accordingly (including `GEPAJobHandle` TS binding changes).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c95ed3bc7e421b52ce41a6d47099dc278cb6849a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tensorzero-optimizers/src/gepa/mod.rs
+++ b/tensorzero-optimizers/src/gepa/mod.rs
@@ -45,11 +45,12 @@ use pareto::{Candidate, ParetoFrontier, is_improvement};
 use validate::{get_uninitialized_variant_configs, validate_examples, validate_gepa_config};
 
 pub use types::{
-    GepaAnalyzeParams, GepaAnalyzeResult, GepaCleanupParams, GepaEvalAndUpdateParams,
-    GepaEvalParentParams, GepaEvalParentResult, GepaIterUpdateResult, GepaIterationParams,
-    GepaIterationResult, GepaMutateParams, GepaMutateResult, GepaSampleParams, GepaSampleResult,
-    GepaSetupParams, GepaSetupResult, GepaSideInfo, GepaToolOutput, GepaToolParams,
-    ParetoCheckpoint, SelectedVariant,
+    GepaAnalyzeContinue, GepaAnalyzeParams, GepaAnalyzeResult, GepaCleanupParams,
+    GepaEvalAndUpdateParams, GepaEvalParentContinue, GepaEvalParentParams, GepaEvalParentResult,
+    GepaIterUpdateResult, GepaIterationParams, GepaIterationResult, GepaMutateContinue,
+    GepaMutateParams, GepaMutateResult, GepaSampleContinue, GepaSampleParams, GepaSampleResult,
+    GepaSetupParams, GepaSetupResult, GepaSideInfo, GepaSkipIteration, GepaToolOutput,
+    GepaToolParams, ParetoCheckpoint, SelectedVariant,
 };
 
 /// A GEPA variant with its name and configuration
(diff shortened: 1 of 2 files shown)


anndvision:
The `run_gepa()`, `run_gepa_setup()`, and `run_gepa_iteration()` functions should take two executor parameters instead of a concrete `&Client`:

```rust
pub async fn run_gepa(
    variant_executor: Arc<dyn EvaluationsInferenceExecutor>,
    judge_executor: Arc<dyn EvaluationsInferenceExecutor>,
    ...
)
```

Similarly, `build_inference_executor()` should return `Arc<dyn EvaluationsInferenceExecutor>` rather than `Client`.

The two-executor architecture is a hard constraint — even though both executors are the same today, this will eventually support separate client-side and server-side executors (e.g., variant inference runs on the user's infra while analysis/mutation runs on our side).

See the `andrew/durable-ize-gepa` branch for the target signatures.

PR #6585, opened 2026-02-25
Add cache.enabled and cache.backend config fields

shuyangli:
Add InferenceCacheBackend enum (Auto, ClickHouse, Valkey) and new fields to ModelInferenceCacheConfig:

- enabled: Option\<bool> - true requires backend, null uses if available, false disables
- backend: InferenceCacheBackend - explicit backend selection (default: auto)

Since we're providing an explicit toggle, in Postgres-primary mode we're going to stop trying to fall back to ClickHouse. Users can explicitly specify clickhouse if they want to.

See https://www.notion.so/tensorzerodotcom/Postgres-backed-TensorZero-config-changes-3107520bbad3804d802bc8d265bc361f?source=copy_link.

Benefits #5691. Will need to document the new config changes soon.

### DIFF
--- a/tensorzero-core/src/config/stored/AGENTS.md
+++ b/tensorzero-core/src/config/stored/AGENTS.md
@@ -1,0 +1,3 @@
+# Stored Config Types
+
+`Stored*Config` types represent configuration that is persisted in the database. They must **not** use `#[serde(deny_unknown_fields)]`, because the database may contain fields added by newer versions of the application after a rollback. Rejecting unknown fields would cause deserialization failures when rolling back to an older version.
--- a/tensorzero-core/src/cache.rs
+++ b/tensorzero-core/src/cache.rs
@@ -61,7 +61,6 @@ impl CacheManager {
             return Ok(Self::disabled());
         }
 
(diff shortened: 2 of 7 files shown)


tensorzero-core/src/config/stored/AGENTS.md
```diff
@@ -0,0 +1,3 @@
+# Stored Config Types
```
GabrielBianconi: Add a `CLAUDE.md` that links to this file. See other examples in the repo.

tensorzero-core/src/config/stored/AGENTS.md
```diff
@@ -0,0 +1,3 @@
...
+
+`Stored*Config` types represent configuration that is persisted in the database. They must **not** use `#[serde(deny_unknown_fields)]`, because the database may contain fields added by newer versions of the application after a rollback. Rejecting unknown fields would cause deserialization failures when rolling back to an older version.
```
GabrielBianconi: How do we test this directory overall? Right now I'm just trusting you that we didn't break anything here... Could be separate PR/etc.

GabrielBianconi (approved):

PR #6671, opened 2026-03-03
Migrate clickhouse-only inference tests to run against both databases

shuyangli:
[#6718](https://github.com/tensorzero/tensorzero/issues/6718)

### DIFF
--- a/tensorzero-core/src/db/inferences.rs
+++ b/tensorzero-core/src/db/inferences.rs
@@ -29,7 +29,7 @@ use crate::serde_util::{
 use crate::stored_inference::{
     StoredChatInferenceDatabase, StoredInferenceDatabase, StoredJsonInference,
 };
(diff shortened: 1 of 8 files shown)


tensorzero-core/src/db/inferences.rs
```diff
@@ -46,8 +46,8 @@ pub(super) struct ClickHouseStoredChatInferenceWithDispreferredOutputs {
     pub output: Vec<ContentBlockChatOutput>,
     #[serde(default)]
     pub dispreferred_outputs: Vec<String>,
-    #[serde(flatten, deserialize_with = "deserialize_tool_info")]
-    pub tool_params: ToolCallConfigDatabaseInsert,
+    #[serde(flatten, deserialize_with = "deserialize_optional_tool_info")]
+    pub tool_params: Option<ToolCallConfigDatabaseInsert>,
```
shuyangli: This was always making it non-optional even for empty cases, which caused failures in the e2e tests

PR #6672, opened 2026-03-03
chore(python): fix incremental typing debt in python client tests and types

Rishabh-git10:
Addresses part of #2521
cc @GabrielBianconi

### Changes
* **`clients/python/tensorzero/types.py`**: Cast JSON deserialization to resolve `reportUnknownVariableType` on the `content` list, and removed unused imports.
* **`clients/python/tests/conftest.py`**: Cast `patch_openai_client` return value to `Awaitable[Any]` to resolve `reportGeneralTypeIssues`.
* **`clients/python/tests/test_embeddings.py`**: Typed the `async_openai_client` fixture as `AsyncOpenAI` to resolve missing parameter types, and updated `pytest.raises` to expect `APIStatusError`.
* **`clients/python/tests/test_extra_content.py`**: Typed the `async_openai_client` fixture, explicitly typed empty lists, and cast custom `messages` arrays to satisfy OpenAI's strict parameter typing.

### DIFF
--- a/clients/python/tests/conftest.py
+++ b/clients/python/tests/conftest.py
@@ -294,7 +294,7 @@ async def async_openai_client(request: FixtureRequest):
         async with AsyncOpenAI(api_key="donotuse") as client:
             await cast(
                 Awaitable[Any],
-                patch_openai_client(
+                patch_openai_client(  # pyright: ignore[reportDeprecated]
                     client,
                     config_file=TEST_CONFIG_FILE,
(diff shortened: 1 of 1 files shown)


GabrielBianconi (approved):
Thanks. This is progress in the right direction. To the extent we can avoid `cast` or `Any` let's do that, but I think we can't fully get rid of them.

Rishabh-git10:
Fixed the `ruff` import sorting.

PR #6769, opened 2026-03-06
fix: only show enabled backends in /health response

### LINKED ISSUE #6750: Hide unused dependencies from `/health`
Postgres, ClickHouse, Valkey, etc. - if some or all are irrelevant for the setup (e.g. Postgres not configured), we should hide them.

koriyoshi2041:
## Summary

Resolves #6750. The `/health` endpoint previously always included all backends (ClickHouse, Postgres, Valkey, Valkey cache) in its response, even when they were disabled. This made the response noisy and potentially confusing for operators.

**Changes:**
- Added `is_clickhouse_enabled()`, `is_postgres_enabled()`, `is_valkey_enabled()` helpers that check concrete connection types
- Changed `health_check_inner()` to accept concrete types instead of `dyn HealthCheckable`, enabling compile-time backend state inspection
- Only includes a backend in the JSON response when it is actually configured
- Replaced `futures::join!` with `tokio::join!` (removes an unnecessary dependency)

**Before:**
```json
{"gateway": "ok", "clickhouse": "ok", "postgres": "ok", "valkey": "ok", "valkey_cache": "ok"}
```

**After** (e.g., only ClickHouse + Postgres configured):
```json
{"gateway": "ok", "clickhouse": "ok", "postgres": "ok"}
```

## Test plan

- [x] Updated existing tests to use concrete types instead of mocks
- [x] Added `test_health_handler_all_disabled` — verifies only "gateway" appears when all backends are disabled
- [x] `cargo check -p tensorzero-core` passes

### DIFF
--- a/crates/tensorzero-core/src/endpoints/status.rs
+++ b/crates/tensorzero-core/src/endpoints/status.rs
@@ -1,13 +1,15 @@
 use crate::{
     db::HealthCheckable,
+    db::clickhouse::{ClickHouseConnectionInfo, clickhouse_client::ClickHouseClientType},
+    db::postgres::PostgresConnectionInfo,
+    db::valkey::ValkeyConnectionInfo,
     utils::gateway::{AppState, AppStateData},
 };
 use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::Json;
-use futures::join;
 use serde::{Deserialize, Serialize};
-use serde_json::{Value, json};
+use serde_json::{Map, Value, json};
 
 pub const TENSORZERO_VERSION: &str = env!("CARGO_PKG_VERSION");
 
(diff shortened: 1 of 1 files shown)


koriyoshi2041:
Could you clarify which tests you expect to fail? The unit tests in `endpoints::status::tests` all pass locally (4/4):

```
test endpoints::status::tests::test_health_handler_all_disabled ... ok
test endpoints::status::tests::should_report_error_for_unhealthy_postgres ... ok
test endpoints::status::tests::test_health_handler_all_enabled_and_healthy ... ok
test endpoints::status::tests::should_report_error_for_unhealthy_clickhouse ... ok
```

Are you referring to specific E2E tests that assert on the health response format? Happy to update those as well.

Issue #4365, opened 2025-11-03
Add optional expiration date to API keys


GabrielBianconi:

GabrielBianconi:
@eibrahim95 Yes, React Router!

PR #6849, opened 2026-03-10
feat: Add expires column and handle expired keys

eibrahim95:
Closes: #4365 
The table now includes an Expires column. Expired keys are treated as inactive, matching the visual style of disabled keys (strikethrough and opacity) and disabling the delete action.

The create_api_key function in napi now accepts an optional RFC3339 timestamp string to set the key's
expiration. The value is parsed and passed to the underlying
authentication module.

The "create api key" modal now accepts an optional expiration date.

Tests were updated, and a new test was added to test the expiration specifically.

### DIFF
--- a/ui/app/components/ui/date-time-picker.tsx
+++ b/ui/app/components/ui/date-time-picker.tsx
@@ -137,7 +137,7 @@ export function DateTimePicker({
           disabled={
             minCalendarDate ? (day) => day < minCalendarDate : undefined
           }
-          initialFocus
+          autoFocus
         />
         <div className="border-t p-3 flex flex-col gap-3 items-center">
           <div className="flex items-center gap-2">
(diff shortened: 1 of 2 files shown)


simeonlee:
Screenshot 2026-03-11 at 11 59 49 AM

Could we play with the spacing a bit? 

- Chevrons are overlapping values in the date selector
- the time icon is a bit close to the edge of the input

simeonlee:
This looks good but picking a date and time is too much UX overhead, good for edge cases though. Can we try a select dropdown with 7 days, 30 days, 90 days, 1 year, Custom and then custom triggers the custom pickers?

simeonlee (changes requested):
Great work! Just a few nits, let me know when you've taken another pass

GabrielBianconi:
@codex review

virajmehta (approved):

PR #6888, opened 2026-03-11
Stop loading configs when rendering evaluations

### LINKED ISSUE #6557: Refactor evaluation config to support defining evaluators at the top level
TODOs for core:
- [x] Evaluations CLI should support these options
- [x] Python library should support these options

TODOs for UI:
- [x] In main evaluation table, "name" should redirect to the run itself
- [x] Evaluation result datapoint page should not take an evaluation name in the route
- [x] Evaluation result datapoint page should have reasonable breadcrumb
- [ ] Sort out evaluation selector so we can select comparable runs across unnamed evaluations (split out to #6859)
    - [ ] Perhaps: for an evaluation run, we know the function from metadata, so find all other evaluation runs where function matches and display them as `run name (variant)` in the dropdown

shuyangli:
Fixes #6557

- Stop loading configs when loading evaluations - we already store all runs in the database, so read those instead. Linking to function configs currently drops the snapshot hash (correct because previously we don't know which version of the function was used in the evaluation anyways), and if we want to do that we can return snapshot hash from the backend when returning eval run metadata.
- Also fixes edge cases when loading multiple runs with an invalid URL: we reject if they don't target the same function, and merge all their metrics. (Instead of taking the first run ID's metadata)

Before:
image

After: image

- Merge metrics across all runs by name, so no metric information is lost when runs have different evaluator sets. This lets us compare different evaluation runs with named and unnamed evaluations for the same function.

 (The two columns are both "exact match" but one is produced by the nested evaluator and on by the flat evaluator; we will add a tooltip to distinguish. They can be different evaluators even though the output is the same, so I don't think we should mix them.)

- Applies the same fix to both `runs.tsx` and `results/route.tsx` loaders.
- Clean up some unused Zod schemas from ui/app/utils/clickhouse/evaluations.ts

Addresses review comments on #6864.

---

> [!NOTE]
> **Medium Risk**
> Changes evaluation results loaders and metric/evaluator mapping logic, which can affect which metrics render and when LLM-judge feedback controls appear. Risk is moderate due to cross-route behavior changes and new validation/merge logic, but scope is limited to evaluation UI data shaping.
>
> **Overview**
> Evaluation pages (`runs` and per-datapoint `results`) now **stop loading server config/snapshots** and instead derive `evaluation_name`, `function_name/type`, and metric metadata solely from `getEvaluationRunMetadata`.
>
> The loaders now **union/merge metadata across all selected runs**, validate all runs target the same function (400 on mismatch), and build metric mappings via new `mergeRunMetrics`; `consolidateEvaluationResults` now takes `metricsConfig` to compute display `evaluator_name` without parsing metric-name strings.
>
> UI components were adjusted accordingly: `BasicInfo` components take explicit `functionName`, `EvaluationTable` no longer depends on `evaluationConfig` and uses `metricsConfig.optimize`, and LLM-judge feedback UI is gated by an `isLlmJudgeEvaluation` flag inferred from result rows.
>
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11e0fe4febb9e4a52327360382bf5f287e373ee2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ui/app/routes/evaluations/EvaluationTable.tsx
+++ b/ui/app/routes/evaluations/EvaluationTable.tsx
@@ -210,10 +210,10 @@ interface EvaluationTableProps {
   selected_evaluation_run_infos: EvaluationRunInfo[];
   evaluation_results: EvaluationResultRow[];
   evaluation_statistics: EvaluationStatistics[];
-  evaluator_names: string[];
+  metric_names: string[];
   evaluation_name: string;
   metricsConfig: Record<string, RunMetricMetadata>;
-  /** Maps evaluator_name → metric_name. */
+  /** Maps full metric_name → short evaluator_name. */
   evaluatorMetricNames: Record<string, string>;
   selectedRows: Map<string, SelectedRowData>;
   setSelectedRows: React.Dispatch<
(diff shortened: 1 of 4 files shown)


shuyangli:
@codex review

PR #6903, opened 2026-03-12
Postgres batch write backpressure

AntoineToussaint:
## Problem

The Postgres write path had two failure modes under load:

1. **Unbounded batch channels** — the original batch writer used unbounded `mpsc` channels. Under sustained load, these grew without limit and lost all unflushed data on process shutdown (~70% data loss in load tests).

2. **WritePool (single-row INSERTs)** — the concurrent worker pool approach used 50 workers each executing single-row INSERTs. At ~850 QPS with 400KB payloads, each inference generated 4 single-row INSERTs (chat/json metadata + data, model metadata + data). Workers couldn't keep up — ~1000 jobs/sec throughput vs ~1700 job submissions. After 60s, the backlog needed ~42s to drain but the test only waited 30s → ~10% data loss.

Neither approach provided backpressure or was suitable for production Postgres deployments.

## Solution: Two-Stage Pipeline

This PR replaces the batch writer with a two-stage pipeline architecture (similar to Vector, Datadog Agent, OpenTelemetry Collector):

```
Rows → bounded input channels → accumulator tasks → bounded flush queue → N flush workers → Postgres
                                  (batch by rows/timeout)                    (bulk INSERTs)
```

### Stage 1: Bounded Input Channels + Accumulators
- Rows are submitted via `try_send` on bounded `mpsc` channels (capacity: `write_queue_capacity`, default 10,000)
- When full, rows are **dropped and logged** (not silently lost)
- One accumulator task per table type collects rows into batches (by `max_rows` or `flush_interval_ms`)

### Stage 2: Concurrent Flush Pool
- Ready batches are submitted as jobs to N flush workers (`flush_concurrency`, default 4)
- Workers execute bulk INSERTs in parallel against the `PgPool`
- Metadata + data INSERTs within each flush run **concurrently** via `tokio::join!` (no FK constraints between tables)

### Backpressure Chain
When Postgres is slow, the full chain applies backpressure:
```
Postgres slow → flush workers blocked → flush queue fills → accumulator blocks on submit
→ input channel fills → try_send fails → caller drops + logs
```

## Configuration

The design philosophy is **zero-config for 99% of users**. Setting `async_writes = true` automatically enables the batch writer with production-ready defaults:

```toml
[gateway]
observability.async_writes = true
```

That's it. No PhD in database tuning required.

### Advanced Tuning (for the remaining 1%)

| Parameter | Default | Description |
|-----------|---------|-------------|
| `write_queue_capacity` | 10,000 | Bounded channel capacity per table type. At 1000 QPS, this is ~10s of buffer. |
| `batch_writes.flush_interval_ms` | 100 | Max time to accumulate before flushing |
| `batch_writes.max_rows` / `max_rows_postgres` | 1,000 | Max rows per batch INSERT |
| `batch_writes.flush_concurrency` | 4 | Number of concurrent flush workers |

### Throughput Math
4 flush workers × 1000-row batches × ~200ms each = **~20,000 rows/sec** per table type. With concurrent metadata+data INSERTs, effectively doubled.

## Other Changes

- **Suppress sqlx statement logging**: sqlx dumps full SQL with thousands of bind params on every query. Replaced with custom slow query logging (1s threshold) that logs table name and row count.
- **Auto-enable batch writes**: When `async_writes = true`, automatically creates a `BatchWritesConfig` with good defaults (no need to explicitly enable `batch_writes`).
- **Stored config compatibility**: `StoredObservabilityConfig` handles `write_queue_capacity` and `flush_concurrency` with serde defaults, with historical snapshot tests.
- **Load test automation**: Added `scripts/load-test/run_postgres_load_test.py` — fully automated load test runner that manages docker compose, compilation, migrations, gateway lifecycle, and produces comparison tables across write-mode variants.

## Load Test Results

At 500 QPS target (local macOS, single gateway instance):

| Metric | Result |
|--------|--------|
| Achieved QPS | 412.6 |
| p99 latency | 4.2ms |
| Error rate | 0.0% |
| DB rows | 24,756 / 24,757 expected |
| **Data loss** | **0%** |

Previously: 10-70% data loss at similar QPS.

## Test Improvements

- **Generic `poll_until_eq` helper**: Extracted a single generic `poll_until_eq<F, Fut>` replacing duplicated `poll_until_count` / `poll_until_model_count` helpers in batch write tests.
- **Replace fixed sleeps with polling**: Replaced fixed `sleep()` calls in `inference_clickhouse.rs` and `feedback.rs` with `poll_for_result` — tests now complete as soon as data is visible rather than waiting an arbitrary fixed duration. Removed a 10s sleep for batch write visibility and multiple 200–500ms sleeps before ClickHouse queries.

## Roadmap: Production / K8s Deployments

For serious on-prem deployments with multiple gateway instances behind a k8s Deployment:

- **Connection pool contention**: N gateway pods × `connection_pool_size` (default 50) connections = N×50 total Postgres connections. Postgres default `max_connections=100` means this needs tuning at scale. Future work: auto-detect or recommend `connection_pool_size = max_connections / expected_replicas`.
- **Flush concurrency scaling**: With many gateway instances, each running 4 flush workers, the total concurrent INSERT pressure on Postgres scales linearly. May want to add adaptive concurrency (measure latency, back off flush workers dynamically).
- **Connection-aware batching**: Future optimization — share flush pool metrics across gateway instances via a lightweight coordination mechanism, or use Postgres advisory locks to coordinate flush scheduling.
- **PgBouncer / connection pooler**: For large deployments (>10 pods), a connection pooler like PgBouncer in transaction mode is the standard solution. The batch writer's bulk INSERTs are transaction-mode friendly since each is a single statement.

## Test plan

- [x] `cargo check --package tensorzero-core --package evaluations`
- [x] `cargo nextest run --package tensorzero-core --lib` (2285 tests pass)
- [x] Historical snapshot tests for `StoredObservabilityConfig` (new fields parse with defaults)
- [x] Load test: 0% data loss at 412 QPS with `batch-throughput` config
- [ ] E2E tests with Postgres batch writes

## Flaky Test Fixes

- **`test_list_datasets_with_pagination`**: Filter pagination queries by a unique `function_name` to avoid interference from concurrent tests creating datasets with overlapping timestamps.
- **`test_streaming_thinking` (GCP Vertex)**: Add `retry_on_rate_limit` and `retry_until_non_rate_limit` helpers with exponential backoff (2s→32s) for 429/RESOURCE_EXHAUSTED errors on follow-up inference calls.
- **`test_insert_inference_evaluation_run_postgres`**: Use `list_evaluation_runs(100, 0)` and find by run_id instead of assuming our run is the most recent (concurrent tests can insert newer runs).
- **`test_image_inference_store_filesystem`**: Replace `.unwrap()` with `.expect()` for better diagnostics.


scripts/load-test/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+services:
+  postgres:
+    image: tensorzero/postgres:14
+    environment:
+      POSTGRES_DB: tensorzero-load-test
+      POSTGRES_PASSWORD: postgres
```
shuyangli: let's use the new tensorzero/postgres image, and also put this in `crates/tensorzero-core/tests/load/postgres-inference-load-test`

docs/gateway/configuration-reference.mdx
```diff
@@ -251,6 +251,7 @@ The `batch_writes` object supports the following fields:
 - `flush_interval_ms` (integer, optional): Maximum time in milliseconds to wait before flushing a batch (default: `100`)
 - `max_rows` (integer, optional): Maximum number of rows to collect before flushing a batch (default: `1000`)
 - `max_rows_postgres` (integer, optional): Maximum number of rows to collect before flushing a Postgres batch (default: `max_rows`)
+- `write_queue_capacity` (integer, optional): Capacity of the bounded input channel per table type. When the channel is full, new rows are dropped and logged instead of buffering without limit. Increase this value if you see `"Postgres batch channel full"` log messages during traffic spikes (default: `10000`)
```
shuyangli: Maybe: "Maximum number of inferences each gateway instance can hold in memory before degrading observability writes. When the write queue is full, we drop new rows and log an error to protect the gateway from out-of-memory crashes. Increase this value if... and monitor your gateway instances' memory usage to determine whether your setting is appropriate." or something to this effect? I want to call out this is for OOMs and reliability. (Can fix in a follow-up.)

shuyangli (approved):

GabrielBianconi:
@virajmehta should review if he hasn't; I'll stamp once he does

PR #6957, opened 2026-03-16
Fix batch inference stuck in Pending when JSONL parsing fails

AntoineToussaint:
## Summary

Fixes a bug where `GET /batch_inference/{batch_id}` returns 400 "Missing batch inference response" indefinitely for GCP Vertex Gemini batch jobs (and potentially other providers). The root cause is a chain of failures when the provider reports a job as completed but the JSONL result file can't be parsed.

## Root cause

When GCP Vertex reports `JOB_STATE_SUCCEEDED`:
1. `collect_finished_batch` reads a JSONL file from GCS
2. `parse_jsonl_batch_file` silently skips all unparseable rows via `continue`, returning an empty `elements` map
3. Empty `elements` → empty `inference_ids` → `get_batch_model_inferences` returns `Ok(vec![])`
4. `batch_model_inferences.first()` returns `None` → `MissingBatchInferenceResponse` error
5. The error propagates via `?` **before** the status update to `Completed` runs (line 882)
6. The batch stays stuck in `Pending` forever — every subsequent poll hits the same failure

## Issues and fixes

| # | Issue | Fix | Test |
|---|-------|-----|------|
| 1 | `parse_jsonl_batch_file` silently returns empty elements when all JSONL rows fail to parse | Return an `InferenceServer` error with row/error counts when `elements` is empty and `total_lines > 0` | 5 unit tests: all valid, partial failure, all fail, empty file, `make_output` failure (`helpers.rs`) |
| 2 | Batch stays stuck in `Pending` when `write_completed_batch_inference` fails | In `write_poll_batch_inference`, catch the error and mark the batch as `Failed` before returning | E2E test `test_write_poll_completed_batch_with_empty_elements_marks_failed` (ClickHouse + Postgres) |
| 3 | Status update failure masks the original root cause error | Use `if let Err(status_err) = ...` instead of `?` — log the status update failure but always return the original error | Covered by the E2E test (if the wrong error propagated, the assertion would fail) |
| 4 | `MissingBatchInferenceResponse` maps to `400 BAD_REQUEST` | Changed to `500 INTERNAL_SERVER_ERROR` — this is a server-side issue, not a client error | Unit test `test_missing_batch_inference_response_is_500` (`error/mod.rs`) |
| 5 | Fire-and-forget `Error::new()` without `let _ =` (confusing intent) | Added `let _ =` and clarifying comment | Style fix, no dedicated test needed |
| 6 | Double iteration over `text.lines()` (count then loop) | Count inline during the loop | Perf fix, covered by existing parse tests |

## Test plan

- [x] All 2273 unit tests pass (`cargo test-unit-fast`)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] 6 new unit tests pass (5 for `parse_jsonl_batch_file`, 1 for status code)
- [x] 1 new E2E test compiles (requires Docker for ClickHouse/Postgres to run)
- [ ] Verify E2E tests pass with `docker compose up` + `cargo nextest`

### DIFF
--- a/crates/tensorzero-core/src/error/mod.rs
+++ b/crates/tensorzero-core/src/error/mod.rs
@@ -2319,6 +2319,25 @@ mod tests {
     }
 
     #[test]
+    fn test_missing_batch_inference_response_is_500() {
+        let error = Error::new(ErrorDetails::MissingBatchInferenceResponse { inference_id: None });
+        assert_eq!(
+            error.status_code(),
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "MissingBatchInferenceResponse should be 500, not 400 — \
+             this is a server-side issue (provider data couldn't be matched to stored inferences)"
+        );
+
+        let error_with_id = Error::new(ErrorDetails::MissingBatchInferenceResponse {
+            inference_id: Some(Uuid::now_v7()),
+        });
+        assert_eq!(
+            error_with_id.status_code(),
+            StatusCode::INTERNAL_SERVER_ERROR,
+        );
+    }
+
+    #[test]
     fn test_extract_from_fatal_stream_error() {
         let error = Error::new(ErrorDetails::FatalStreamError {
             message: "stream died".to_string(),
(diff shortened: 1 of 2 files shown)


virajmehta (approved):
This seems like a good change, however, are we sure this is the only way we could get the error that was reported? I didn't go through the whole code path.

PR #7046, opened 2026-03-23
Expose function/evaluator to Autopilot

shuyangli:
Step towards #7013


shuyangli:
now we're eagerly validating

PR #7093, opened 2026-03-27
feat: Add VariantStatistics materialized view

### LINKED ISSUE #6980: Create a materialized view `VariantStatistics` that includes the same statistics as `ModelProviderStatistics`
It will need `function_name`, `variant_name`, `minute`, `processing_time_ms_quantiles`, `ttft_ms_quantiles`, `total_input_tokens`, `total_output_tokens`, `count`, `total_cost`, and the new columns for KV cache reads + writes. We'll need to join on `InferenceById` in ClickHouse to get the variant names associated with each model inference and also aggregate per-inference first to get the response_time from the ChatInference and the JsonInference tables as well as make sure the counts are correct.

In Postgres this should be more straightforward.

AntoineToussaint:
## Summary

- **Denormalizes `function_name` and `variant_name` onto `ModelInference`** (ClickHouse migration 0053, Postgres migration). This eliminates the need for JOINs with `InferenceById` in materialized views, avoiding ordering issues in batch mode where the JOIN target might not yet exist when the MV fires.
- Adds `VariantStatistics` aggregation tables that group metrics by `(function_name, variant_name, minute)` — the variant-level counterpart of `ModelProviderStatistics`
- **ClickHouse** (migration 0053): Adds denormalization columns, then creates `AggregatingMergeTree` table fed by three materialized views — `VariantStatisticsModelView` (tokens/cost from `ModelInference` using denormalized columns, no JOIN), `VariantStatisticsChatView` (latency quantiles + count from `ChatInference`), and `VariantStatisticsJsonView` (same from `JsonInference`)
- **Postgres**: `variant_statistics` table with `refresh_variant_statistics_incremental()` function, plus denormalization columns on `model_inferences`
- No backfill of historical data — too risky for production (OOM on large ClickHouse JOINs, row locks on Postgres). MVs capture all new data going forward. Historical data can be backfilled manually if needed.
- Adds `prepare_variant_statistics()` test helper for both backends
- Adds E2E aggregation tests verifying token sums, cost aggregation (with NULL exclusion), inference count, and `count_with_cost` across multiple model inferences per inference

Closes #6980

## Test plan

- [x] `cargo check --all-targets --all-features`
- [x] `cargo fmt`
- [x] `cargo test-unit-fast` (12 pre-existing failures, no new ones)
- [x] CI green (all checks pass)
- [ ] E2E: `test_variant_statistics_aggregation_clickhouse` — inserts chat + model inferences, verifies aggregated row
- [ ] E2E: `test_variant_statistics_aggregation_postgres` — same test on Postgres backend
- [ ] ClickHouse migration e2e: migration 0053 applies cleanly on fresh DB

### DIFF
--- a/crates/tensorzero-core/src/db/clickhouse/migration_manager/migrations/migration_0053.rs
+++ b/crates/tensorzero-core/src/db/clickhouse/migration_manager/migrations/migration_0053.rs
@@ -89,12 +89,12 @@ impl Migration for Migration0053<'_> {
             ))
             .await?;
 
-        // Note: we do NOT backfill existing ModelInference rows because ClickHouse
-        // does not support correlated subqueries in ALTER TABLE UPDATE. Instead:
+        // Historical ModelInference rows are NOT backfilled (ClickHouse does not
+        // support correlated subqueries in ALTER TABLE UPDATE). This is fine:
         // - New rows: application code populates function_name/variant_name on insert.
-        // - VariantStatistics backfill (Phase 3): JOINs ChatInference/JsonInference
-        //   for historical data, so denormalized columns aren't needed for old rows.
-        // - The MV's `function_name != ''` filter correctly excludes unbackfilled rows.
+        // - The MV's `function_name != ''` filter excludes old unbackfilled rows.
+        // - Historical VariantStatistics data is also not backfilled; only new
+        //   inserts going forward are captured by the MVs.
 
         // ── Phase 2: Create VariantStatistics table and materialized views ──
 
--- a/crates/tensorzero-core/src/inference/types/mod.rs
+++ b/crates/tensorzero-core/src/inference/types/mod.rs
@@ -1551,7 +1551,7 @@ impl ModelInferenceResponseWithMetadata {
 
 impl StoredModelInference {
     /// Create a new StoredModelInference from a runtime ModelInferenceResponseWithMetadata.
-    /// Used when inserting into ClickHouse.
+    /// Used when inserting into the database.
     pub async fn new(
         result: ModelInferenceResponseWithMetadata,
         inference_id: Uuid,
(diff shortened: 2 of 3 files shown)


crates/tensorzero-core/src/db/clickhouse/migration_manager/migrations/migration_0053.rs
```diff
@@ -0,0 +1,366 @@
...
+                .response
+                .trim()
+                != "0";
+
+            if !has_data {
+                tracing::info!(
+                    "Skipping `VariantStatistics` backfill: no data in `ModelInference`"
+                );
+                return Ok(());
+            }
```
virajmehta: we should not do this, just run the commands and they'll grab zero rows if we're in that case

crates/tensorzero-core/src/db/clickhouse/migration_manager/migrations/migration_0053.rs
```diff
@@ -0,0 +1,366 @@
...
+            MIGRATION_ID,
+        )
+        .await?
+        {
+            self.clickhouse
+                .run_query_synchronous_no_params(format!(
+                    "ALTER TABLE ModelInference{on_cluster_name} ADD COLUMN IF NOT EXISTS variant_name LowCardinality(String) DEFAULT ''"
+                ))
+                .await?;
+        }
```
virajmehta: so the historical data is going to be empty? this seems quite bad to have this be just '' in old data

crates/tensorzero-core/src/db/postgres/migrations/20260330000000_model_inferences_denormalize.sql
virajmehta: I do not want ragged data

virajmehta (changes requested):

crates/tensorzero-core/src/db/postgres/migrations/20260330000000_model_inferences_denormalize.sql
AntoineToussaint: Removed this migration entirely.

virajmehta (dismissed):

PR #7163, opened 2026-04-01
fix: reduce live test flakiness

AntoineToussaint:
## Summary

- **Increase e2e slow-timeout from 30s to 60s**: nextest was killing live tests at 60s (`period=30s, terminate-after=2`). Real provider APIs (OpenAI, Anthropic, etc.) can take longer than that, especially on shared CI runners. Now tests terminate at 120s instead, matching what several provider-specific overrides already use.

This should fix the recurring merge queue failures on `live-tests / live-tests (batch_writes: false, database: clickhouse)`.

## Test plan

- [x] Verify the nextest config change is correct
- [ ] Observe that live tests pass more reliably in the merge queue after merging


virajmehta (approved):
LGTM not sure why the test is failing here

PR #7278, opened 2026-04-13
Support hotswapping database connections

### LINKED ISSUE #7255: Support hot-swapping ClickHouse/Postgres connection info
## Context

`clickhouse_connection_info` and `postgres_connection_info` are currently excluded from the swappable `RuntimeDependencies` bundle because they interfere with batch writes on shutdown.

During shutdown, `GatewayHandle::drop` grabs the batch writer join handles from the current `RuntimeDependencies`, drops `app_state`, then waits for the batch writers to drain. If these connections were swappable, a config swap could replace the live connections mid-shutdown, causing the old batch writers to be dropped before they finish flushing.

## What needs to happen

To safely support hot-swapping these connections, the shutdown logic needs to be updated to:

1. Collect batch writer handles from **all** generations of connections (current + any in-flight swapped-out ones), not just the latest snapshot.
2. Wait for all of them to drain before exiting.

Alternatively, we could prevent hot-swaps from completing during shutdown by checking the `CancellationToken` inside `swap_config`.

## Current workaround

These two connection infos are held as fixed fields on `SwappableAppStateData` and are not swappable via `prepare_config_swap`/`swap_config`.

shuyangli:
Fixes #7255

We remove both RuntimeDependencies and DeferredShutdown, and move all live fields into LiveState. Then, impl Drop on LiveState so we track the batch writer drain tasks for shutdown.

- Delete RuntimeDependencies struct; move all its fields directly into LiveState
- LiveState is no longer Clone (ArcSwap doesn't require it, and Clone+Drop would spawn duplicate drain tasks)
- LiveState::Drop spawns batch-writer drain tasks onto the shared deferred_tasks tracker, so all generations (not just the latest) are drained on shutdown
- Add clickhouse_url and postgres_url to ConnectionUrls; prepare_config_swap now creates fresh ClickHouse/Postgres connections on each config swap using the stored URLs
- Remove DeferredShutdown (its drain logic is now in LiveState::Drop)
- Add unit test: test_all_live_state_generations_drained_on_shutdown

---

> [!NOTE]
> **Medium Risk**
> Changes the gateway’s hot-swap and shutdown behavior by recreating DB connections on config swaps and moving batch-writer draining into `LiveState::drop`, which could impact in-flight requests or shutdown ordering if mis-handled.
> 
> **Overview**
> Enables **hot-swapping database connections** during config reloads by moving all runtime dependencies (HTTP client, ClickHouse/Postgres/Valkey connection info, cache manager, and `primary_datastore`) directly into the `LiveState` stored in `ArcSwap`.
> 
> Config swaps now rebuild fresh ClickHouse/Postgres/Valkey connections from stored `ConnectionUrls` and atomically replace the entire `LiveState`. Shutdown/drain logic is shifted to a `Drop` impl on `LiveState`, which spawns batch-writer join-handle waits onto the shared `deferred_tasks` so **all prior generations** are drained, not only the latest.
> 
> Adds a unit test (`test_all_live_state_generations_drained_on_shutdown`) to verify old-generation batch writers are awaited after a swap + gateway drop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6aa0e9681fb8d11cc52d66b0d0e4aa6b2f84c8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/tensorzero-core/src/utils/gateway.rs
+++ b/crates/tensorzero-core/src/utils/gateway.rs
@@ -132,8 +132,14 @@ impl Drop for GatewayHandle {
 
 /// All hot-swappable state for a running gateway.
 /// Stored inside `Arc<ArcSwap<LiveState>>` so config swaps are atomic.
-/// Not `Clone` — it holds a `Drop` impl that spawns batch-writer drain tasks,
-/// and `ArcSwap` only requires `Send + Sync`, not `Clone`.
+///
+/// Not `Clone` — if cloned, when dropped we will wait on the same database connection's batch writer multiple times,
+/// which works but is logically wrong.
+///
+/// This only spawns tasks that *wait* for shutdown - the shutdown happens automatically
+/// once all outstanding `ClickHouseConnectionInfo` and `PostgresConnectionInfo` handles are dropped.
+/// It's therefore safe for us to hand out cloned `ClickHouseConnectionInfo` and `PostgresConnectionInfo` handles
+/// from this struct.
 struct LiveState {
     config: Arc<Config>,
     runtime_overlay: Arc<RuntimeOverlay>,
@@ -145,6 +151,7 @@ struct LiveState {
     valkey_cache_connection_info: ValkeyConnectionInfo,
     cache_manager: CacheManager,
     primary_datastore: PrimaryDatastore,
+    rate_limiting_manager: Arc<RateLimitingManager>,
     /// Shared with `SwappableAppStateData.deferred_tasks`.
     /// When this `LiveState` generation is dropped (because a config swap replaced it),
     /// we spawn tasks here to drain the batch writers before the gateway fully shuts down.
@@ -154,8 +161,23 @@ struct LiveState {
 
 impl Drop for LiveState {
     fn drop(&mut self) {
-        if let Some(clickhouse_handle) = self.clickhouse_connection_info.batcher_join_handle() {
-            self.deferred_tasks.spawn(async move {
+        // Destructure all fields so the compiler errors if a new field is added without
+        // considering whether it needs drain/shutdown logic here.
+        let Self {
+            config: _,
+            runtime_overlay: _,
+            http_client: _,
+            clickhouse_connection_info,
+            postgres_connection_info,
+            valkey_connection_info: _,
+            valkey_cache_connection_info: _,
+            cache_manager: _,
+            primary_datastore: _,
+            rate_limiting_manager: _,
+            deferred_tasks,
+        } = self;
+        if let Some(clickhouse_handle) = clickhouse_connection_info.batcher_join_handle() {
+            deferred_tasks.spawn(async move {
                 tracing::info!("Waiting for ClickHouse batch writer to finish");
                 if let Err(e) = clickhouse_handle.await {
                     tracing::error!("Error in batch writer: {e}");
@@ -163,8 +185,8 @@ impl Drop for LiveState {
                 tracing::info!("ClickHouse batch writer finished");
             });
         }
-        if let Some(postgres_handle) = self.postgres_connection_info.batcher_join_handle() {
-            self.deferred_tasks.spawn(async move {
+        if let Some(postgres_handle) = postgres_connection_info.batcher_join_handle() {
+            deferred_tasks.spawn(async move {
                 tracing::info!("Waiting for Postgres batch writer to finish");
                 if let Err(e) = postgres_handle.await {
                     tracing::error!("Error in batch writer: {e}");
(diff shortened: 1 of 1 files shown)


shuyangli:
@bugbot review

crates/tensorzero-core/src/utils/gateway.rs
```diff
@@ -130,101 +130,59 @@ impl Drop for GatewayHandle {
...
-    fn disabled(config: &Config, primary_datastore: PrimaryDatastore) -> Self {
-        Self {
-            http_client: config.http_client.clone(),
-            valkey_connection_info: ValkeyConnectionInfo::Disabled,
-            valkey_cache_connection_info: ValkeyConnectionInfo::Disabled,
-            cache_manager: CacheManager::disabled(),
-            primary_datastore,
+impl Drop for LiveState {
+    fn drop(&mut self) {
+        if let Some(clickhouse_handle) = self.clickhouse_connection_info.batcher_join_handle() {
```
Aaron1011: Can you do an explicit `let LiveState { clickhouse_connection_info, postgres_connection_info, <other fields> } = self` here, so that we get a compiler error if we add new fields?

crates/tensorzero-core/src/utils/gateway.rs
```diff
@@ -243,43 +201,6 @@ impl SwappableConfig {
     }
 }
 
-/// Holds state that needs to have tasks spawned onto `deferred_tasks`
-/// when dropped.
-/// Currently, we use this to ensure that we wait for the batch writer handles to finish
-/// when the gateway shuts down.
-/// This only spawns tasks that *wait* for shutdown - the shutdown happens automatically
```
Aaron1011, line 250: Can you copy this part of the comment to `LiveState`?

crates/tensorzero-core/src/utils/gateway.rs
```diff
@@ -433,23 +372,30 @@ impl SwappableAppStateData {
         self.live_state.store(Arc::new(LiveState {
             config: prepared.config,
             runtime_overlay: prepared.runtime_overlay,
-            runtime_dependencies: prepared.runtime_dependencies,
+            http_client: prepared.http_client,
```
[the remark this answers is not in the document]
shuyangli: done

crates/tensorzero-core/src/utils/gateway.rs
shuyangli: done

crates/tensorzero-core/src/utils/gateway.rs
shuyangli: done

shuyangli:
@codex review

crates/tensorzero-core/src/utils/gateway.rs
```diff
@@ -130,101 +130,81 @@ impl Drop for GatewayHandle {
...
-            primary_datastore,
+            valkey_connection_info: _,
+            valkey_cache_connection_info: _,
+            cache_manager: _,
+            primary_datastore: _,
+            rate_limiting_manager: _,
+            deferred_tasks,
+        } = self;
+        if let Some(clickhouse_handle) = clickhouse_connection_info.batcher_join_handle() {
+            deferred_tasks.spawn(async move {
```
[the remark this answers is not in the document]
shuyangli: that's embedded only which doesn't get swaps, i think this is fine

crates/tensorzero-core/src/utils/gateway.rs
```diff
@@ -405,99 +345,143 @@ impl SwappableAppStateData {
...
-                &config,
-                self.connection_urls.as_ref(),
-                &self.deferred_shutdown.clickhouse_connection_info,
-                &self.deferred_shutdown.postgres_connection_info,
-            )
-            .await?,
-        );
+
+        let current_live_state = self.live_state.load_full();
+        // When a URL is stored, rebuild the connection fresh (picks up any URL changes and runs
```
Aaron1011, line 350: Does this mean that reloading the config will always throw away our existing connection pool (even in the common case where the database config settings don't change)? If not, can you add a comment?

shuyangli: yeah it does; we are not diffing the URLs. I'm actually thinking maybe we should just require a gateway restart if the database connections are changed, since it'll be annoying to conditionally swap things out by diffing the configs. should we just punt it?

Aaron1011: Yeah, that makes sense to me. I think it's also far more likely for someone to hotswap the database config by accident then to actually want to change the target database on a running gateway. We could still do most of this refactor, though.

Aaron1011:
Separately, I realized the hotswapping postgres could lead to our normal connection limit being exceeded - we'll have two pools constructed at once, each with the same limit. If gateway is under heavy load when a postgres url hostswap occurs, we could have twice the normal maximum number of connections until all of the connections on the old pool finished executing.

crates/tensorzero-core/src/utils/gateway.rs
shuyangli: I'll do some the refactor in another PR to get rid of RuntimeDependencies, but still keep Clickhouse/Postgres.

PR #7284, opened 2026-04-14
Flatten RuntimeDependencies into LiveState

shuyangli:
Pure refactor only to clean up some unnecessary structs

### DIFF
--- a/crates/tensorzero-core/src/utils/gateway.rs
+++ b/crates/tensorzero-core/src/utils/gateway.rs
@@ -149,6 +149,12 @@ struct LiveState {
     valkey_cache_connection_info: ValkeyConnectionInfo,
     cache_manager: CacheManager,
     primary_datastore: PrimaryDatastore,
+    /// Token pool manager for rate limiting pre-borrowing.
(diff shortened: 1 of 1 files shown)


crates/tensorzero-core/src/utils/gateway.rs
```diff
@@ -430,46 +387,47 @@ impl SwappableAppStateData {
     /// This is infallible; all fallible work is done up front in
     /// [`Self::prepare_config_swap`].
     pub fn swap_config(&self, prepared: PreparedConfigSwap) {
+        self.rate_limiting_manager
```
Aaron1011: Is there any harm in storing this inside `LiveState` (and just re-using the old one when constructing the `PreparedConfigSwap`)? If we have two different `ArcSwap` fields, then a request might see a partially-updated config + rate-limiting manager, depending the execution timing

shuyangli: yeah good call, done

Aaron1011 (approved):

PR #7293, opened 2026-04-15
Fix feedback table layout for long metric names and many tags

### LINKED ISSUE #6437: Feedback table on inference detail page breaks with long metric names or many tags
## Problem

On the inference detail page, the feedback table layout breaks when rows have long metric names or many tags:

1. **Long metric names** (e.g. `tensorzero::evaluation_name::extract_entities_accuracy::evaluator_name::accuracy`) push the Metric column extremely wide, squeezing other columns
2. **Many tags** stack vertically in the Tags column, making the row extremely tall
3. The ID and Value columns get compressed — ID is truncated to just a few chars, Value gets squeezed

The result is a row that's both very wide and very tall, with poor use of space.

## Expected

- Long metric names should truncate or wrap gracefully (e.g. with `truncate` + tooltip on hover)
- Tags column should have a max height or collapse after N tags (e.g. show 3 + "+7 more")
- Column widths should be better balanced so ID and Value aren't squeezed

ianliuy:
## Summary

Fix the feedback table layout that breaks when rows have long metric names or many tags.

This is a minimal CSS-only fix for the 3 layout problems described in #6437. Unlike #6617 (which was a full 31-file redesign), this PR targets only the specific layout issues in 2 source files with no data flow or architecture changes.

Fixes #6437

### Metric column (`FeedbackTable.tsx`)

**Before:** `<TableCell className="flex items-center gap-2">` applied `display: flex` directly on the `<td>` (non-standard — the only place in the codebase doing this), with no width constraint on the metric name span.

**After:** Flex layout moved to an inner `<div>` wrapper. Cell gets `max-w-[250px]`, metric name gets `truncate` + Radix `Tooltip` for full name on hover, badges get `shrink-0` to prevent squishing.

### Tags column (`FeedbackTable.tsx`)

Added `max-w-[200px]` constraint on the Tags cell to prevent unbounded expansion.

### Tag collapse (`TagsBadges.tsx`)

**Before:** All tags rendered with `flex-wrap`, stacking vertically when numerous.

**After:** Limited to `MAX_VISIBLE_TAGS = 3`. Extra tags collapse into a "+N more" badge with a `Tooltip` showing the hidden tags on hover.

### Tests & stories

- 6 new unit tests for tag collapse logic (`TagsBadges.test.tsx`)
- 3 new Storybook regression stories: `WithLongMetricName`, `WithManyTags`, `WithLongMetricNameAndManyTags`

## Test plan

- [x] All 11 unit tests pass (5 existing + 6 new)
- [x] ESLint passes on all changed files
- [ ] Verify long metric names truncate with tooltip on hover
- [ ] Verify tags collapse to "+N more" when >3 tags
- [ ] Verify Storybook stories render correctly
- [ ] Verify inference detail, episode detail, and preview sheet all display feedback table properly

cc @shuyangli @GabrielBianconi

### DIFF
--- a/ui/app/components/feedback/FeedbackTable.stories.tsx
+++ b/ui/app/components/feedback/FeedbackTable.stories.tsx
@@ -408,3 +408,161 @@ export const WithVariousTags: Story = {
     },
   },
 };
+
+// Regression stories for issue #6437:
+// Layout breaks with long metric names or many tags
+
+export const WithLongMetricName: Story = {
+  args: {
+    feedback: [
+      {
+        type: "float",
+        id: makeOrderedUuid(10),
+        target_id: TARGET_ID,
+        metric_name:
+          "tensorzero::evaluation_name::extract_entities_accuracy::evaluator_name::accuracy",
+        value: 0.92,
+        tags: { user_id: "123" },
+        timestamp: "2024-03-20T10:00:00Z",
+      },
+      {
+        type: "boolean",
+        id: makeOrderedUuid(9),
+        target_id: TARGET_ID,
+        metric_name:
+          "tensorzero::evaluation_name::summarization_quality_check::evaluator_name::hallucination_detection",
+        value: false,
+        tags: {},
+        timestamp: "2024-03-20T10:01:00Z",
+      },
+      {
+        type: "float",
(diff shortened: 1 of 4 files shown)


ianliuy:

